PR #1, opened 2017-08-13
Deploy using apex/up

mccutchen:
Basic support for deploying via https://github.com/apex/up:

screen shot 2017-08-12 at 8 16 52 pm

screen shot 2017-08-12 at 8 17 44 pm

Currently deployed here:
https://wg7c5oygsl.execute-api.us-west-2.amazonaws.com/development/

This deployment doesn't actually work that well, because `go-httpbin` doesn't know how to handle the `/development` URL prefix that seems to be unavoidable until a custom domain is configured.

TODO:
- [x] custom domain w/ TLS
- [x] get rid of the `/{stage}` URL prefix (I _think_ [using a custom domain is required to make this happen]?)

[1]: http://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html

### DIFF
--- a/.upignore
+++ b/.upignore
@@ -1,0 +1,6 @@
+.git
+cmd
+dist
+httpbin
+static
+Dockerfile
--- a/.upignore
+++ b/.upignore
@@ -1,6 +1,4 @@
-.git
-cmd
-dist
-httpbin
-static
-Dockerfile
+*
+!server
+!main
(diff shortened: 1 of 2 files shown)


mccutchen:
Now deployed to
- https://httpbingo.org/
- https://stage.httpbingo.org/

PR #2, opened 2018-03-20
Fix body parsing with longer form content-type headers

na--:
This fixes `parseBody()` when the `Content-Type` header is something like `application/x-www-form-urlencoded; charset=utf-8`.


mccutchen:
Awesome, thanks @na--!

PR #3, opened 2018-03-21
Fix deflate encoded responses

na--:
RFC 2616 (HTTP/1.1) says in the ["3.5 Content Codings" section](https://tools.ietf.org/html/rfc2616#section-3.5) this about deflate:
> The "zlib" format defined in RFC 1950 [31] in combination with the "deflate" compression mechanism described in RFC 1951 [29].

The [zlib package](https://golang.org/pkg/compress/zlib/#pkg-overview) from Go's standard library imlpements RFC 1950 and it also [uses](https://golang.org/src/compress/zlib/writer.go#L15) the [flate package](https://golang.org/pkg/compress/flate/#pkg-overview) internally, which implements RFC 1951.

I'm not sure if you maintain httpbingo.org, but opening https://httpbingo.org/deflate with Firefox or Chrome produces an error, I assume because of this. For some reason `curl -v --compressed https://httpbingo.org/deflate` works though...

PR #4, opened 2018-03-21
Add support for HEAD and fix the OPTIONS HTTP method

na--:
Currently an `OPTIONS` request returns `405 Method Not Allowed` error and `HEAD` is not supported at all.

Edit: I just saw that this was branched from the [other PR](https://github.com/mccutchen/go-httpbin/pull/3), if you want I can cherry pick and create a new PR from `master`, but if you merge the other one first it should be fine as it is.

### DIFF
--- a/httpbin/middleware.go
+++ b/httpbin/middleware.go
@@ -4,8 +4,32 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/http/httptest"
 	"time"
 )
+
+func optionsAndHead(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case "OPTIONS":
+			w.Header().Set("Access-Control-Allow-Methods", "GET, POST, HEAD, PUT, DELETE, PATCH, OPTIONS")
+			w.Header().Set("Access-Control-Max-Age", "3600")
+			if r.Header.Get("Access-Control-Request-Headers") != "" {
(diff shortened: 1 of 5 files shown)


mccutchen:
Would you mind rebasing these changes now that #3 has landed?

httpbin/middleware.go
```diff
@@ -17,13 +41,6 @@ func cors(h http.Handler) http.Handler {
 		respHeader.Set("Access-Control-Allow-Origin", origin)
 		respHeader.Set("Access-Control-Allow-Credentials", "true")
 
-		if r.Method == "OPTIONS" {
```
mccutchen: Sorry, but it's not obvious to me why this code needed to be moved into a different middleware invoked before this `cors` middleware.  Would you mind explaining?

na--: I moved it in another handler just for clarity and because I couldn't figure out how to name a function that handles both CORS and `HEAD`/`OPTIONS` request methods :slightly_smiling_face:  There's no technical reason they can't be in the same handler, but the same could probably be said for `logger()` and `limitRequestSize()`. 

Also, due to the way these handlers are composed, the execution order is `cors`, `optionsAndHead`, `logger`,  `limitRequestSize` and finally the `mux`. But now that I think about it, it shuld probably be `cors`, `logger`, `optionsAndHead`,  `limitRequestSize`, otherwise the log will treat `HEAD` requests as `GET`.

na--:
I'll rebase and create a new PR from master that also includes the slight fix in middleware ordering I noted in my other comment

PR #5, opened 2018-03-22
Add support for HEAD and fix the OPTIONS HTTP method

na--:
This is a rebased version of https://github.com/mccutchen/go-httpbin/pull/4 with a fixed middleware order.

At the moment `OPTIONS` requests return `405 Method Not Allowed` error and `HEAD` is not supported at all.

### DIFF
--- a/httpbin/middleware.go
+++ b/httpbin/middleware.go
@@ -4,8 +4,32 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/http/httptest"
 	"time"
 )
+
+func optionsAndHead(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case "OPTIONS":
+			w.Header().Set("Access-Control-Allow-Methods", "GET, POST, HEAD, PUT, DELETE, PATCH, OPTIONS")
+			w.Header().Set("Access-Control-Max-Age", "3600")
+			if r.Header.Get("Access-Control-Request-Headers") != "" {
+				w.Header().Set("Access-Control-Allow-Headers", r.Header.Get("Access-Control-Request-Headers"))
+			}
+			w.WriteHeader(200)
+		case "HEAD":
+			rwRec := httptest.NewRecorder()
+			r.Method = "GET"
+			h.ServeHTTP(rwRec, r)
+
+			copyHeader(w.Header(), rwRec.Header())
+			w.WriteHeader(rwRec.Code)
+		default:
+			h.ServeHTTP(w, r)
+		}
+	})
+}
 
 func cors(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -17,13 +41,6 @@ func cors(h http.Handler) http.Handler {
 		respHeader.Set("Access-Control-Allow-Origin", origin)
 		respHeader.Set("Access-Control-Allow-Credentials", "true")
 
-		if r.Method == "OPTIONS" {
-			respHeader.Set("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, PATCH, OPTIONS")
-			respHeader.Set("Access-Control-Max-Age", "3600")
-			if r.Header.Get("Access-Control-Request-Headers") != "" {
-				respHeader.Set("Access-Control-Allow-Headers", r.Header.Get("Access-Control-Request-Headers"))
-			}
-		}
 		h.ServeHTTP(w, r)
 	})
 }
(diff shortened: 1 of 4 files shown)


na--:
Ah, actually the logger still does not properly log `HEAD` requests, because it first passes the buck to the next handler so it can measure the request time and after that logs the information, I'll fix it in a bit.

na--:
Accidentally closed the PR by pushing the wrong button... Anyway, I chose to copy the request URI as well as the method just in case it's changed by some of the other handlers downstream in the future.

httpbin/middleware.go
```diff
@@ -17,13 +41,6 @@ func cors(h http.Handler) http.Handler {
 		respHeader.Set("Access-Control-Allow-Origin", origin)
 		respHeader.Set("Access-Control-Allow-Credentials", "true")
 
-		if r.Method == "OPTIONS" {
```
mccutchen: Revisiting the conversation in https://github.com/mccutchen/go-httpbin/pull/4#discussion_r176261518:

>>  @mccutchen Sorry, but it's not obvious to me why this code needed to be moved into a different middleware invoked before this cors middleware. Would you mind explaining?
> 
> @na-- I moved it in another handler just for clarity and because I couldn't figure out how to name a function that handles both CORS and HEAD/OPTIONS request methods 🙂 There's no technical reason they can't be in the same handler, but the same could probably be said for logger() and limitRequestSize().

I understand why there is a new middleware for special casing `HEAD` requests, but what is still not obvious to me is why this specific `OPTIONS`-handling code from the `cors` middleware needs to be moved from here into the new `HEAD`-handling middleware.

na--: Basically my reasoning was that `CORS` is its own thing, it doesn't have anything to do with `OPTIONS`, it's applicable for all HTTP methods. While `HEAD` and `OPTIONS` are kind of the special-case HTTP methods. `HEAD` always doesn't have a body, `OPTIONS` often doesn't have one as well. Both have much more in common than either has with CORS handling.

na--: So the `OPTIONS` code doesn't **need** to be moved, but if I have a new middleware for handling `HEAD` methods, it makes much more sense to stick the `OPTIONS` handling there rather than with the CORS handling.

mccutchen: But … `OPTIONS` requests are explicitly a part of CORS, and the headers originally being set here (that have been moved into the mixed `optionsAndHead` middleware) are CORS-specific headers.

My preference would be to drop this change and simplify the new middleware down to just handling `HEAD` requests, if you're up for it.

_[ time passes ]_

Actually, I finally do see one of the reasons for this change:  With the current code, an `OPTIONS` request will end up invoking the underlying handler and sending back an unnecessary response body, right?  Maybe what we need is to extract a common `discardBody` middleware or helper function that the `cors` and `head` middlewares could use?

na--: Ah, my bad, I totally missed that the emitted headers in `OPTIONS` were actually CORS related... And yes, without this patch `OPTIONS` invokes the underlying handler, which besides being unnecessary also made most requests return `405 Method Not Allowed` as a status.

Not sure if we need to over-complicate things with another middleware though. Besides, `OPTIONS` and `HEAD` discard the body in very different ways - `HEAD` actually needs to know what its length is, while options can just return without any execution of the chained handlers after ir. 

I think it would be simpler to just combine `cors` and `optionsAndHead`, I just have to think of a name... :smile:  But the log should be executed before both handlers, so it can correctly identify `HEAD` requests and actually detect `OPTIONS` ones.

mccutchen: Hmmm, I love a good naming debate!  My first idea for a name for a combined middleware is `preflight`, which is (mostly) accurate for `OPTIONS` but less so for `HEAD`.

What about `metaRequests`, since `OPTIONS` and `HEAD` requests can kinda-sorta be thought of as "requests about requests"?

na--: I like `metaRequests`, will push a new commit with it in a bit

na--: Done

httpbin/middleware.go
```diff
@@ -92,10 +103,11 @@ func (mw *metaResponseWriter) Size() int {
 
 func logger(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		reqMethod, reqURI := r.Method, r.URL.RequestURI()
```
mccutchen: Capturing `reqMethod` is necessary here because our new `metaRequests` middleware will modify the method for `HEAD` requests, right?

na--: Yes, but also because something else could conceivably modify the method or URL in the downstream handlers as well, if not now then in the future. It seemed simpler and more future-proof for the `logger` to save the method and URI just in case, and it's much more efficient than cloning the whole request in the downstream headers.

mccutchen: Cool, makes sense!

PR #14, opened 2019-03-03
Adding support for specifying host and https

saurabh-hirani:
go-httpbin is great and I use it for demos locally. Thanks for building it. 

I ran into a requirement of using https with go-httpbin and added support for the same by adding -host, -https-cert and -https-key options. Updated README with command line and docker examples.

### DIFF
--- a/cmd/go-httpbin/main.go
+++ b/cmd/go-httpbin/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"log"
@@ -13,16 +14,23 @@ import (
 	"github.com/mccutchen/go-httpbin/httpbin"
 )
 
+const defaultHost = "0.0.0.0"
 const defaultPort = 8080
 
 var (
-	port        int
-	maxBodySize int64
-	maxDuration time.Duration
+	host          string
+	port          int
+	maxBodySize   int64
+	maxDuration   time.Duration
+	httpsCertFile string
+	httpsKeyFile  string
 )
 
 func main() {
+	flag.StringVar(&host, "host", defaultHost, "Host to listen on")
 	flag.IntVar(&port, "port", defaultPort, "Port to listen on")
+	flag.StringVar(&httpsCertFile, "https-cert-file", "", "HTTPS Server certificate file")
+	flag.StringVar(&httpsKeyFile, "https-key-file", "", "HTTPS Server private key file")
 	flag.Int64Var(&maxBodySize, "max-body-size", httpbin.DefaultMaxBodySize, "Maximum size of request or response, in bytes")
 	flag.DurationVar(&maxDuration, "max-duration", httpbin.DefaultMaxDuration, "Maximum duration a response may take")
 	flag.Parse()
@@ -47,6 +55,9 @@ func main() {
 			os.Exit(1)
 		}
 	}
+	if host == defaultHost && os.Getenv("HOST") != "" {
+		host = os.Getenv("HOST")
+	}
 	if port == defaultPort && os.Getenv("PORT") != "" {
 		port, err = strconv.Atoi(os.Getenv("PORT"))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mccutchen:
@saurabh-hirani this is really cool, thanks!  Sorry for the slow reply, I was out vacation for a little while.

A couple of high level comments/questions:

 1. My preference would be to make the `-https-cert` and `-https-key` params represent the _path_ to the cert and key files on disk, rather than the actual values themselves.  While allowing the values to be passed in directly is convenient (esp. in a containerized environment), it [goes against best practices](https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/).

 2. Should we consider more explicitly encouraging modern TLS config best practices, as outlined in [Achieving a Perfect SSL Labs Score with Go][perfect-score] and [So you want to expose Go on the Internet][exposing-go]?  This may be getting a bit too extreme, given how security-critical httpbin is likely **not** to be.  If nothing else, I'd probably personally benefit from having a working example of such best practices code in place for me to reference in the future!

[perfect-score]: https://blog.bracebin.com/achieving-perfect-ssl-labs-score-with-go
[exposing-go]: https://blog.cloudflare.com/exposing-go-on-the-internet/#cryptotls

mccutchen:
P.S. If we do decide to beef up the TLS config out of the box, a more fully worked example is available here:
https://github.com/denji/golang-tls/blob/master/README.md#perfect-ssl-labs-score-with-go

saurabh-hirani:
Hi @mccutchen 

Thanks for the feedback. 

1. I did go with the file path initially but while testing the docker image I realized that I would've to update the Dockerfile to mount the cert volumes also - hence I went the env var route which was easier but as I understand that it would be better to use file paths. I will change the code to do the same.

2. Including TLS best practices seems like a good idea. I will explore that. 

I will see if I can make time over the weekend or before to make the changes and update the thread.

saurabh-hirani:
@mccutchen  - added support to use -https-cert-file and -https-key-file instead of taking them from env vars.

saurabh-hirani:
@mccutchen I explored the above links for getting stronger ssl score. But I can't test SSL labs score for localhost and I feel that making the requirements rigid might cause issues for end users who may or may not be inclined to resolve them if their client calls using curl fail (e.g. supporting specific set of ciphers which may not be supported by the client due to older version). I haven't played with letsencrypt enough to do this the right way. Can we use this PR for adding https support and maybe track making it more secure in a separate request?

cmd/go-httpbin/main.go
```diff
@@ -64,7 +82,25 @@ func main() {
...
+
+	if httpsCertFile != "" && httpsKeyFile != "" {
+		cert, err := tls.LoadX509KeyPair(httpsCertFile, httpsKeyFile)
+		if err != nil {
+			log.Fatal("Failed to generate https key pair: ", err)
+		}
+		server.TLSConfig = &tls.Config{
+			Certificates: []tls.Certificate{cert},
+		}
+		logger.Printf("go-httpbin listening on %s with https", listenAddr)
```
mccutchen: Minor, but for usability's sake maybe

```go
logger.Printf("go-httpbin listening on https://%s", listenAddr)
```

would be better?  If nothing else, it would allow folks whose terminals will automagically recognize URLs to click on the link directly?

(Same suggestion would apply to the plain HTTP case below, too!)

saurabh-hirani: Sure - will do that and update the thread.

saurabh-hirani: @mccutchen updated the message and squashed.

mccutchen:
@saurabh-hirani this is awesome, thanks so much!

saurabh-hirani:
@mccutchen - this was fun. Thanks for the guidance. Can you please recreate and upload the docker image so that it reflects the https support?

mccutchen:
D’oh, yeah, I’ll do that later tonight or tomorrow morning. Thanks for the
reminder, I need to figure it out why automated builds aren’t working!

mccutchen:
@saurabh-hirani docker image `mccutchen/go-httpbin:v2.1.1` pushed!

PR #16, opened 2019-06-02
Add support for /uuid

saurabh-hirani:
- Added code and test cases for /uuid endpoint
- Updated httpbin.go to handle ListenAndServe errors - useful because in the current state, the code silently fails in case the server is rerun with the same host and port combination.

### DIFF
--- a/cmd/go-httpbin/main.go
+++ b/cmd/go-httpbin/main.go
@@ -89,18 +89,22 @@ func main() {
 		Handler: h.Handler(),
 	}
 
+	var listenErr error
 	if httpsCertFile != "" && httpsKeyFile != "" {
 		cert, err := tls.LoadX509KeyPair(httpsCertFile, httpsKeyFile)
 		if err != nil {
-			log.Fatal("Failed to generate https key pair: ", err)
+			logger.Fatal("Failed to generate https key pair: ", err)
 		}
 		server.TLSConfig = &tls.Config{
 			Certificates: []tls.Certificate{cert},
 		}
 		logger.Printf("go-httpbin listening on https://%s", listenAddr)
-		server.ListenAndServeTLS("", "")
+		listenErr = server.ListenAndServeTLS("", "")
 	} else {
 		logger.Printf("go-httpbin listening on http://%s", listenAddr)
-		server.ListenAndServe()
+		listenErr = server.ListenAndServe()
+	}
+	if listenErr != nil {
+		logger.Fatalf("Failed to listen: %s", listenErr)
 	}
 }
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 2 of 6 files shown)


saurabh-hirani:
The Travis build failed with 

```
httpbin/handlers.go:15:2: cannot find package "github.com/google/uuid" in any of:
	/home/travis/.gimme/versions/go1.9.linux.amd64/src/github.com/google/uuid (from $GOROOT)
	/home/travis/gopath/src/github.com/google/uuid (from $GOPATH)
``` 

Will check on how to fix it.

saurabh-hirani:
@mccutchen  - I can update .travis.yaml to include **github.com/google/uuid** and update go.mod in the same manner as https://github.com/skyrocknroll/go-mod-example/blob/master/go.mod but this will break the  zero-dependency outside stdlib feature of go-httpbin. Wanted your input on the same before going further. 

httpbin.org which uses https://github.com/postmanlabs/httpbin/ is also depending on a few third party modules as per https://github.com/postmanlabs/httpbin/blob/master/setup.py#L38 - but let me know your thoughts. We can write a local uuid generation function but I'd rather offload that work if someone has done a better job at it.

mccutchen:
> I can update .travis.yaml to include **github.com/google/uuid** and update go.mod in the same manner as https://github.com/skyrocknroll/go-mod-example/blob/master/go.mod but this will break the zero-dependency outside stdlib feature of go-httpbin. Wanted your input on the same before going further.

My preference would be to maintain the lack of external dependencies for this project for as long as possible.  How would you feel about providing a "good enough for government work" uuid4 function built out of a `[16]byte` slice and https://golang.org/pkg/crypto/rand/#Read?

saurabh-hirani:
@mccutchen  - sure - will try it out and update this thread.

saurabh-hirani:
@mccutchen - updated as per discussion.

httpbin/helpers.go
```diff
@@ -230,3 +232,26 @@ func sha1hash(input string) string {
 	h := sha1.New()
 	return fmt.Sprintf("%x", h.Sum([]byte(input)))
 }
+
+func isValidUUIDv4(uuid string) error {
```
mccutchen: Let's put this function in the test file, to help make it obvious that it's only used as a test helper.

cmd/go-httpbin/main.go
```diff
@@ -97,10 +98,13 @@ func main() {
 		server.TLSConfig = &tls.Config{
 			Certificates: []tls.Certificate{cert},
 		}
-		logger.Printf("go-httpbin listening on https://%s", listenAddr)
-		server.ListenAndServeTLS("", "")
+		log.Printf("go-httpbin listening on https://%s", listenAddr)
```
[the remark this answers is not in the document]
saurabh-hirani: No - I missed out on that. Will fix it.

httpbin/helpers.go
saurabh-hirani: Makes sense. Will do.

cmd/go-httpbin/main.go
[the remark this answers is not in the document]
saurabh-hirani: Done.

httpbin/helpers.go
saurabh-hirani: Done.

mccutchen:
This is great, thanks so much @saurabh-hirani!

mccutchen:
I’ll push up tagged docker images and deploy to httpbingo.org in the morning!

PR #17, opened 2019-06-11
Deploy to App Engine

mccutchen:
Here we move https://httpbingo.org over to Google App Engine, which can now run standard issue Go HTTP servers without requiring any App Engine-specific libraries/customizations.

Giving up on [Now](https://zeit.co/now) because V2 can no longer deploy arbitrary docker containers and their support for Go requires [jumping through odd hoops](https://zeit.co/docs/v2/deployments/official-builders/go-now-go/) that I don't want to impact this codebase.

(Gave up on [Up](https://up.docs.apex.sh/) a while back, because I didn't particularly like the complexity/ mystery/magic required to contort an application into AWS's API Gateway.)


mccutchen:
Quick little load test:

```
$ echo "GET https://httpbingo.org/ip" | vegeta attack -rate=500 -duration=120s | vegeta report
Requests      [total, rate]            60000, 500.01
Duration      [total, attack, wait]    2m0.146692174s, 1m59.997999921s, 148.692253ms
Latencies     [mean, 50, 95, 99, max]  140.578254ms, 132.998145ms, 210.142153ms, 256.064539ms, 1.057514031s
Bytes In      [total, mean]            2280000, 38.00
Bytes Out     [total, mean]            0, 0.00
Success       [ratio]                  100.00%
Status Codes  [code:count]             200:60000  
Error Set:
```

(For context, when deployed on Now, the majority of those requests are rejected with `HTTP 429` for exceeding some unknown rate limit.)

PR #21, opened 2019-07-02
Adding /bearer support

saurabh-hirani:
- Not documented in http://httpbin.org/legacy but present in the code as https://github.com/postmanlabs/httpbin/blob/master/httpbin/core.py#L999


httpbin/handlers_test.go
```diff
@@ -2276,6 +2276,69 @@ func TestJSON(t *testing.T) {
...
+	}
+	for _, test := range errorTests {
+		t.Run("error"+test.authorizationHeader, func(t *testing.T) {
+			r, _ := http.NewRequest("GET", requestURL, nil)
+			if test.authorizationHeader != "" {
+				r.Header.Set("Authorization", test.authorizationHeader)
+			}
+			w := httptest.NewRecorder()
+			handler.ServeHTTP(w, r)
+			assertHeader(t, w, "WWW-Authenticate", "Bearer")
```
mccutchen: Let's also `assertStatusCode()` in these tests to ensure we're sending back the expected `http.StatusUnauthorized` response

httpbin/httpbin.go
```diff
@@ -93,6 +93,11 @@ type HTTPBin struct {
 	Observer Observer
 }
 
+type bearerResponse struct {
+	Authenticated bool   `json:"authenticated"`
+	Token         string `json:"token"`
+}
```
mccutchen: Sorry for the _extreme_ nit pick, but let's move this response type up above the `type HTTPBin struct {}` declaration so that it's grouped with the other response types.

httpbin/handlers_test.go
saurabh-hirani: Done.

httpbin/httpbin.go
saurabh-hirani: Makes sense. Done. I would've done that too :)

PR #22, opened 2019-07-02
Support cmd/go-httpbin (👍) and cmd/go_httpbin (👎)

mccutchen:
In #17, we switched to deploying httpbingo.org on Google App Engine. To meet their naming requirements, we had to rename the `cmd/go-httpbin` dir to `cmd/go_httpbin`, but this broke any existing uses of, e.g.,

    go get github.com/mccutchen/go-httpbin/cmd/go-httpbin

as suggested in the README and as used in various places (including my employer).

Here's a dumb ass fix for that dumb ass problem; I'm not happy about it, but it seems to work:  Add a new `github.com/mccutchen/go-httpbin/cmd/maincmd` package that is consumed by the (now minimal) `github.com/mccutchen/go-httpbin/cmd/{go-httpbin,go_httpbin}` main packages.


mccutchen:
Gonna merge this and see what breaks 🙈

PR #20, opened 2019-06-22
Adding /json support and sorting index.html links for readability

saurabh-hirani:
- Added /json endpoint support. Not documented in http://httpbin.org/legacy but present in the code as https://github.com/postmanlabs/httpbin/blob/master/httpbin/core.py#L1749 and usable as http://httpbin.org/json
- Sorted the links in index.html alphabetically for readability. 
- Followed the code of /html https://github.com/mccutchen/go-httpbin/blob/master/httpbin/handlers.go#L589 for consistency in implementation.

PR #23, opened 2019-07-03
Clean up index markup & re-sort links

mccutchen:
A small follow-up to #20, which (in part) sorted the links on the index page:
 - drop unnecessary markup
 - normalize markup to ensure that the list of links can be directly sorted by path
   - e.g. `<del><a href="/brotli">...</a></del>` → `<a href="/brotli"><del>...</del></a>`

cc @saurabh-hirani


saurabh-hirani:
Makes sense.

PR #25, opened 2019-07-03
Get code coverage back up

mccutchen:
Two things here:
 1. Test all base64 error cases
 2. Just let `uuidv4()` panic if it can't read enough random data (which is a legitimately exceptional circumstance)


mccutchen:
Let's see how this newfangled codecov.io integration works in practice

PR #26, opened 2019-10-03
Shut down gracefully

mccutchen:
Ensure that go-httpbin allows any in-flight requests to finish before shutting down.

I'm not sure how best to unit test this, so for now I've just tested it manually like so:

1. In one shell, `make run`
2. In another shell, `curl -s -v localhost:8080/delay/10 >/dev/null` to make a 10 second request
3. In a third shell, `kill -s TERM $(ps aux | grep go-httpbin | grep -v grep | awk '{print $2}')` to tell the server to shut down

This results in these logs from the server, showing that the request was allowed to finish before it exited:

```
time="2019-10-03T08:22:28.5216" msg="received signal terminated, shutdown started"
time="2019-10-03T08:22:35.2511" status=200 method="GET" uri="/delay/10" size_bytes=204 duration_ms=10000.60
time="2019-10-03T08:22:35.5227" msg="shutdown finished"
```

and the `curl` command did get back a 200 OK response:

```
$ curl -s -v localhost:8080/delay/10 >/dev/null
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /delay/10 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Origin: *
< Content-Length: 204
< Content-Type: application/json; encoding=utf-8
< Date: Thu, 03 Oct 2019 15:22:35 GMT
< Connection: close
< 
{ [204 bytes data]
* Closing connection 0
```

### DIFF
--- a/cmd/maincmd/main.go
+++ b/cmd/maincmd/main.go
@@ -2,7 +2,6 @@ package maincmd
 
 import (
 	"context"
-	"crypto/tls"
 	"flag"
 	"fmt"
 	"log"
@@ -78,6 +77,16 @@ func Main() {
 		httpsKeyFile = os.Getenv("HTTPS_KEY_FILE")
 	}
 
+	var serveTLS bool
+	if httpsCertFile != "" || httpsKeyFile != "" {
+		serveTLS = true
+		if httpsCertFile == "" || httpsKeyFile == "" {
+			fmt.Fprintf(os.Stderr, "Error: https cert and key must both be provided\n\n")
+			flag.Usage()
+			os.Exit(1)
+		}
+	}
+
 	logger := log.New(os.Stderr, "", 0)
 
 	// A hacky log helper function to ensure that shutdown messages are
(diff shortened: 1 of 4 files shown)


cmd/maincmd/main.go
```diff
@@ -90,22 +104,48 @@ func Main() {
...
+
+	go func() {
+		sig := <-shutdownCh
+		serverLog("shutdown started by signal: %s", sig)
+
+		shutdownTimeout := maxDuration + 1*time.Second
+		ctx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+
+		server.SetKeepAlivesEnabled(false)
```
Integralist: @mccutchen I noticed you didn't do this with SSO, was there a reason for that?

mccutchen: @Integralist good eye! I had it in my original SSO PR (that one followed this one), but Justin rightly [asked me to explain it](https://github.com/buzzfeed/sso/pull/262#discussion_r340689367).  To save you a click, here's my response:

> I can't [explain it]!  I _thought_ I had copied this from [the stdlib's Shutdown example](https://golang.org/pkg/net/http/#Server.Shutdown), but looking at it again I do not see this called.
> 
> Went ahead and dropped this.

Just never followed up here.

PR #27, opened 2020-01-25
- 1 fix and allow faster "stream-bytes" endpoint

kgersen:
- fix: original code limits stream-bytes & bytes endpoints to 100 * 1024 instead of MaxBodySize
- enhancement: stream-bytes sends the same chunk over and over instead of computing all chunks. greatly improve performance on high speed networking (tested a 10Gbps ok). 
- bug/todo:  -max-duration is not enforced 
- enh/todo: parse units for argument (10KB, 5MB, 12GB, for instance) using https://github.com/alecthomas/units/issues or equivalent.

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -664,26 +664,26 @@ func (h *HTTPBin) ETag(w http.ResponseWriter, r *http.Request) {
 
 // Bytes returns N random bytes generated with an optional seed
 func (h *HTTPBin) Bytes(w http.ResponseWriter, r *http.Request) {
-	handleBytes(w, r, false)
+	h.handleBytes(w, r, false)
 }
 
 // StreamBytes streams N random bytes generated with an optional seed in chunks
 // of a given size.
 func (h *HTTPBin) StreamBytes(w http.ResponseWriter, r *http.Request) {
-	handleBytes(w, r, true)
+	h.handleBytes(w, r, true)
 }
 
 // handleBytes consolidates the logic for validating input params of the Bytes
 // and StreamBytes endpoints and knows how to write the response in chunks if
 // streaming is true.
-func handleBytes(w http.ResponseWriter, r *http.Request, streaming bool) {
+func (h *HTTPBin) handleBytes(w http.ResponseWriter, r *http.Request, streaming bool) {
 	parts := strings.Split(r.URL.Path, "/")
 	if len(parts) != 3 {
 		http.Error(w, "Not found", http.StatusNotFound)
 		return
 	}
 
-	numBytes, err := strconv.Atoi(parts[2])
+	numBytes, err := strconv.ParseInt(parts[2], 10, 64);
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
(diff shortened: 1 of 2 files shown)


httpbin/handlers.go
```diff
@@ -740,19 +737,33 @@ func handleBytes(w http.ResponseWriter, r *http.Request, streaming bool) {
 	rng := rand.New(src)
 
 	w.Header().Set("Content-Type", "application/octet-stream")
+	w.Header().Set("Content-Length", strconv.FormatInt(numBytes, 10))
```
mccutchen: Setting the content type prevents us from streaming the body back to the client for the /stream-bytes/ endpoint, because it prevents the stdlib from automatically adding the `Transfer-Encoding: chunked` header.

The easiest way I know to quickly verify this, since I could not figure out how to actually unit test it (see the comment in the deleted test code below, which should also be restored), is to use telnet:

```
telnet localhost 8080
```

Type this in and hit enter a couple of times:

```
GET /stream-bytes/1024?chunk_size=10 HTTP/1.1
Host: foo
```

You should see a `Transfer-Encoding: chunked` header and a bunch of 10 byte chunks instead of a single 1024 byte chunk.

httpbin/handlers.go
```diff
@@ -740,19 +737,33 @@ func handleBytes(w http.ResponseWriter, r *http.Request, streaming bool) {
...
 	}
-	if len(chunk) > 0 {
-		write(chunk)
+
+	f := w.(http.Flusher)
+	// send the buffer as many times as needed
+	var numChunk = numBytes / chunkSize
+	// fmt.Println("numChunk :", numChunk)
+	for i = 0; i < numChunk; i++ {
+		w.Write(chunk)
```
mccutchen: WIth this change, it appears we now generate only chunkSize random bytes and write those same bytes to the response for every chunk.  Let's maintain the original behavior, where numBytes random bytes are generated and written to the response.

httpbin/handlers_test.go
```diff
@@ -1989,21 +1989,17 @@ func TestStreamBytes(t *testing.T) {
...
 			handler.ServeHTTP(w, r)
 
-			// TODO: The stdlib seems to automagically unchunk these responses
-			// and I'm not quite sure how to test this:
-			//
-			//     assertHeader(t, w, "Transfer-Encoding", "chunked")
-			//
-			// Instead, we assert that we got no Content-Length header, which
-			// is an indication that the Go stdlib streamed the response.
-			assertHeader(t, w, "Content-Length", "")
```
mccutchen: As mentioned above, I'd rather maintain this (admittedly odd) test, or figure out a way to actually test the streaming behavior.

httpbin/handlers.go
kgersen: Oh I see a misunderstanding on my part here. 
For me 'stream-bytes' is about "streaming bytes" (since there is "application/octet-stream") and not about 'Transfer-Encoding: chunked'. The chunk part is to control the flushing not the 'chunked' behavior as defined in https://tools.ietf.org/html/rfc7230#section-4.1 (aka add a size in hex and carriage returns between each chunks which is what stdlib does hence your 10 byte chunks).

So I removed 'Transfer-Encoding: chunked' because it's usually not really compatible with 'application/octet-stream' (I think it's open to interpretation here since httpbin does it).

 may be there should be 2 different endpoints.

httpbin/handlers.go
kgersen: The whole point for me is to have a base code for performance measurement. Randomizing every bytes just kill the performance on very high speed network and render this project useless for me.

I started this because I need a http server to get/post for performance measurements. Instead of starting from scratch I noticed your project and decided to try building upon. That the whole point of this PR, to see if our goal are compatible or not.

I think now it's not the good approach. Since your projet can be used as a lib, I think it's better for me to build my own project and 'speed measurement' endpoints and add your endpoints as a lib (if possible).

You can forget this PR then. Sorry about that.

mccutchen: > I started this because I need a http server to get/post for performance measurements. Instead of starting from scratch I noticed your project and decided to try building upon. That the whole point of this PR, to see if our goal are compatible or not.

Gotcha!  Peak performance has never been a specific goal for this project, but I would not be opposed to adding a new endpoint to stream an arbitrary number of fixed bytes.  I could even talk myself into it being compatible with this /stream-bytes endpoint as an additional `?bytes=<specific chunk to stream back>` parameter to allow the client to control the actual bytes that will be streamed.

> I think now it's not the good approach. Since your projet can be used as a lib, I think it's better for me to build my own project and 'speed measurement' endpoints and add your endpoints as a lib (if possible).

That approach also seems pretty reasonable.  If you do go down this road, please let me know if you run into any trouble "embedding" go-httpbin's endpoints into your more purpose-built server, I'd love to know where the pain points are and whether they can be addressed.

> You can forget this PR then. Sorry about that.

No worries, thanks for opening it!

PR #31, opened 2020-04-18
Experiment with GitHub Actions

mccutchen:
Let's see if we can:
- [x] replace Travis w/ GitHub Actions (for some reason Travis is no longer reporting build statuses)
- [x] still integrate with codecov.io

Bonus points for:
- [ ] automated staging deploys for all pull requests whose tests pass
- [ ] automated production deploys on merge

PR #32, opened 2020-04-18
Continuous delivery

mccutchen:
Here we integrate automated deploys into our new GitHub Actions workflows (#31), with these goals:
- [x] automatically deploy to staging for all branches with passing tests
- [x] automatically deploy to production on merge


mccutchen:
Let's see whether the production deploy and staging cleanup workflows actually work 🤞🤞 🤞

PR #33, opened 2020-04-21
Fix staging deploy cleanup workflow

mccutchen:
In an unsurprising turn of events, there was a mistake in the cleanup workflow from #32.


mccutchen:
🙏 🙏 🙏

PR #36, opened 2020-08-10
Allow serving non-index pages via plain HTTP (fixes #35)

### LINKED ISSUE #35: httpbingo.org: HTTP redirects to HTTPS
httpbingo.org currently redirects all requests form HTTP to HTTPS, making explicit HTTP tests impossible. For maximum compatibility with httpbin.org it should answer in exactly the protocol that was requested.

mccutchen:
Per #35, httpbingo.org requiring https for all requests actually reduces its utility as a general purpose http testing tool.  Here we tweak the App Engine deployment configuration to require https only for requests to the index page, but not to any other resources.

My thought is that it's still worthwhile to use https by default, especially for users who end up at httpbingo.org in a web browser, but we'll see whether this actually works the way I'm hoping it will.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -3,9 +3,14 @@ runtime: go112
 
 main: ./cmd/go_httpbin
 
-# redirect everything to https
 handlers:
-  - url: /.*
+  # Always redirect index requests to https
+  - url: /
     script: auto
     secure: always
     redirect_http_response_code: 301
+
+  # Allow requests for any other resources via either http or https
+  - url: /.+
+    script: auto
+    secure: optional
--- a/app.yaml
+++ b/app.yaml
@@ -1,5 +1,5 @@
 ---
-runtime: go112
+runtime: go113
 
 main: ./cmd/go_httpbin
 


mccutchen:
Strike one:

```
ERROR: (gcloud.app.deploy) An error occurred while parsing file: [/home/runner/work/go-httpbin/go-httpbin/app.yaml]
Unable to assign value '^/$' to attribute 'url':
Value '^/$' for url does not match expression '^(?:(?!\^)/.*|\..*|(\(.).*(?!\$).)$'
  in "/home/runner/work/go-httpbin/go-httpbin/app.yaml", line 8, column 10
```

mccutchen:
Looks to be working as expected on staging:

```bash
# index page is still redirected to https
$ curl -v http://pr36-dot-httpbingo.uc.r.appspot.com/ 2>&1 | grep -e '< HTTP' -e 'Location'
< HTTP/1.1 301 Moved Permanently
< Location: https://pr36-dot-httpbingo.uc.r.appspot.com/

# other resources are not
$ curl -v http://pr36-dot-httpbingo.uc.r.appspot.com/headers 2>&1 | grep -e '< HTTP' -e 'Location'
< HTTP/1.1 200 OK
```

Issue #40, opened 2020-09-28
/drip endpoint seems broken?


StefanKarpinski:
Finding this project very useful! Super easy to deploy and use for CI. One endpoint that seems to be broken is the `/drip` endpoint. When I fetch it with libcurl, the progress callback shows that I only get one response packet. By comparison, when I use `https://httpbin.org/drip` I get ten packets of a byte each.

mccutchen:
Hi, thanks for the report!  To clarify, are you testing against https://httpbingo.org or against an instance of go-httpbin running locally?

I did a little digging, and it turns out that this is definitely broken on https://httpbingo.org, because Google App Engine (where it's currently deployed) [explicitly disallows streaming responses](https://cloud.google.com/appengine/docs/standard/go/how-requests-are-handled):

> App Engine does not support streaming responses where data is sent in incremental chunks to the client while a request is being processed. All data from your code is collected as described above and sent as a single HTTP response.

That's a real shame, and breaks a few of the current endpoints in the same way.  I guess I've got to find another place to host this little thing!

The good news is that, as far as I can tell, the `/drip` endpoint should work with a local instance of go-httpbin, but I'd love to know if that's not what you're seeing.  If this is not working as expected against a local instance, can you provide a quick way to reproduce?

The [tests for this endpoint](https://github.com/mccutchen/go-httpbin/blob/40a126261600a23a71417529cef4c2d43631ee82/httpbin/handlers_test.go#L1545-L1565) are a little weird to write (IIRC the golang standard library http client makes it hard to directly test because it "helpfully" handles the chunked transfer encoding before it gets to my code).

The curl command line client seems to get a streaming response locally, as shown here (note the `--no-buffer` flag to prevent curl from buffering the whole response):

```
curl -v --no-buffer 'http://localhost:8080/drip?duration=10&numbytes=10'
```

![Untitled2](https://user-images.githubusercontent.com/57938/94493490-c4aaba00-01ba-11eb-84ec-f2c6b6b2e17c.gif)

mccutchen:
(Opened #41 to track the task of finding a new home for httpbingo.org)

StefanKarpinski:
I had tried both https://httpbingo.org and https://httpbingo.julialang.org, which @staticfloat just setup for the Julia project so we don't tax your servers with our CI and so that we're more insulated from network issues. Both had the issue but it may be the same problem. @staticfloat, do you think it's the same issue for our setup?

mccutchen:
[...]

PR #43, opened 2020-10-01
Make /drip more compatible with original httpbin's implementation

mccutchen:
This is the rest of the fix for #40, to bring our `/drip` implementation more in line with that of httpbin.org's:

- Set `Content-Length` header
- Match default parameter values
- Send status code before any delay

cc @StefanKarpinski


StefanKarpinski:
Looks good to me!

mccutchen:
I'm going to land these changes now, which will have the unfortunate side effect of re-breaking the `/drip` endpoint's streaming response once it's deployed to httpbingo.org.

Will continue to evaluate good options for low-maintenance, low-cost places to host this application in the meantime.

PR #55, opened 2021-05-05
Use go:embed directive to embed static files

empire:
Remove go-bindata and use go1.16 embed package instead.


mccutchen:
Thanks for the contribution!  I'd love to get rid of go-bindata, but I think I'd rather hold off on breaking backwards compatibility with older golang versions at least until the next release.

A related change I'd like to make is stamping a go-modules compatible version, which I think at this point would require bumping up to v3.  Maybe that's actually a reasonable opportunity to introduce a change like this.

Let me think about it a bit, but this is very helpful.  Thanks again!

mccutchen:
Side note:  Looks like the tests are still run against go 1.15 by default, so I guess we'll need to explicitly ask for 1.16 here:
https://github.com/mccutchen/go-httpbin/blob/934a2a788c8869244ec668d30c6acaee64ef0f97/.github/workflows/continuous_delivery.yaml#L21-L22

mccutchen:
See https://github.com/mccutchen/go-httpbin/pull/56, which we can land separately from these `//go:embed` changes.

empire:
@mccutchen Thanks for your comments, I think we can support both the older and newer go versions together until the release of go1.17. Thanks for your previous changes, as the CI didn't run I wasn't able to get the error properly. Thanks for the fix.

empire:
> See #56, which we can land separately from these `//go:embed` changes.

Based on #56, Isn't it better to drop the go1.13 version and just support two recent versions of go?

empire:
> Thanks for the contribution! I'd love to get rid of go-bindata, but I think I'd rather hold off on breaking backwards compatibility with older golang versions at least until the next release.

The latest code is compatible with older go versions. Could you please check it? Thanks

mccutchen:
> The latest code is compatible with older go versions.

I'm afraid I don't see the reason to use both go-bindata and the new `embed.FS` approach in the same codebase, separated by build flags.  Is there some advantage to this approach for 1.16+ users that is not obvious to me?

empire:
Actually, there is not a strong reason behind it, I am ok to close this PR until go1.17 becomes available.

mccutchen:
I appreciate your patience, @empire!  I'm gonna close this for now, but I'll re-open it and make sure you get credit for these changes when it's time to land them.  Thanks again for the contribution!

empire:
@mccutchen, go1.17 was released. Are you agree with using the embed package?

Issue #63, opened 2021-07-19
can you push multi-arch images to docker-hub?


danbf:
![200 (2)](https://user-images.githubusercontent.com/5041307/126233432-014a1f34-5b9d-4973-a19d-a00e11fb35aa.gif)

do you think you can start pushing multi-arch images to docker-hub ala'
```
docker buildx build --platform linux/amd64,linux/arm64,linux/arm/v7
```

https://www.docker.com/blog/multi-arch-images/

mccutchen:
👋 hey @danbf!

![name](https://user-images.githubusercontent.com/57938/126234132-e07169c0-0f12-4ece-a6eb-eeabf868c22d.gif)

I could probably do something like that … I need to automate release builds on merge, might as well push up multi-arch builds while I'm at it.  Do you have a particular need for each of those architectures?  (Left to my own devices, I'd probably just stick to `linux/amd64` and `linux/arm64` as the most likely architectures.)

danbf:
glad to see you are still finding time for maintaining this 😄 

it's hackweek and i'm doing rig on arm, which is turning out to be surprisingly easy, and almost certainly going to go into production. for that `linux/amd64` and `linux/arm64` seems to satisfy aws ecs's mix of regular and graviton's.  `linux/arm/v7` seems to be mostly used by the pi folk.

no rush.  should be pretty straightforward since your upstream is already multi-arch.

mccutchen:
@danbf while I figure out how best to automate this, I just manually built and pushed a multi-arch build to the tag `mccutchen/go-httpbin:v2.2.2-multiarch`:

<img width="914" alt="Screen Shot 2021-07-21 at 10 29 23 AM" src="https://user-images.githubusercontent.com/57938/126506058-21fbc1e3-6b1d-4883-876e-399a6bc2efb2.png">

Lemme know if that works for you!

danbf:
![giphy (6)](https://user-images.githubusercontent.com/5041307/126573268-250efc73-6688-44a4-b16c-620231c7503b.gif)

double awesome with a cherry on top.

danbf:
since buildx needs some setup, i do this on my end for mult-arch's

ensure i cleared out the buildx builder in case it got interrupted before
```
docker buildx rm <builder-name>
```
then:
```
docker buildx create --name <builder-name>
docker buildx use <builder-name>
docker buildx build --tag <remote_tag> . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
```
[...]

PR #64, opened 2021-07-21
Build docker images for amd64 and arm64

mccutchen:
Initial implementation of #63.


Makefile
```diff
@@ -142,10 +142,10 @@ gcloud-auth:
 # docker images
 # =============================================================================
 image:
-	DOCKER_BUILDKIT=1 docker build -t $(DOCKER_TAG_DOCKERHUB) .
+	docker buildx build --platform linux/amd64,linux/arm64 -t $(DOCKER_TAG_DOCKERHUB) .
 
-imagepush: image
-	docker push $(DOCKER_TAG_DOCKERHUB)
+imagepush:
+	docker buildx build --push --platform linux/amd64,linux/arm64 -t $(DOCKER_TAG_DOCKERHUB) .
```
danbf, line 148: i think this needs to be re-jiggered based on how buildx works and how it needs a `builder` defined for it to use:

this has worked for me although the first step of ensuring any orphaned buildx builders could be improved be parsing the output of `docker buildx ls`

since buildx needs some setup, i do this on my end for multi-arch's:

ensure i cleared out the buildx builder in case it got interrupted before
```
docker buildx rm <builder-name>
```
then:
```
docker buildx create --name <builder-name>
docker buildx use <builder-name>
docker buildx build --tag <remote_tag> . --platform linux/amd64,linux/arm64,linux/arm/v7 --push
```
then clear out the buildx builder so it's ready for the next time i run it
```
docker buildx rm <builder-name>
```

danbf: i got it running in `sso` like this: https://github.com/buzzfeed/sso/pull/319

danbf:
i'm thinking this might be what's needed https://github.com/mccutchen/go-httpbin/pull/65

have to look a little more.  the biggest change is that with buildx, we don't store the image locally and then push, we need to build and push in one operation.  so we might need to build twice, once for local stuff, tests, etc, and the other build and push up to dockerhub.

.github/workflows/continuous_delivery.yaml
```diff
@@ -18,13 +18,16 @@ jobs:
...
+
+    - name: Setup go
       uses: actions/setup-go@v2
       with:
         go-version: '1.16'
 
-    - name: Checkout
-      uses: actions/checkout@v2
+    - name: Set up docker buildx
+      uses: docker/setup-buildx-action@v1
```
danbf, line 30: first off, you are doing some sophisticated stuff in here between the workflow and Makefile. that said since we are pushing images here i don't see we need to make any changes to the workflow.

however if we do want to start pushing here, i think i found a guide:

https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md

Makefile
```diff
@@ -142,10 +142,10 @@ gcloud-auth:
 # docker images
 # =============================================================================
 image:
-	DOCKER_BUILDKIT=1 docker build -t $(DOCKER_TAG_DOCKERHUB) .
+	docker buildx build --platform linux/amd64,linux/arm64 -t $(DOCKER_TAG_DOCKERHUB) .
```
danbf, line 145: i don't think we need to change this, just the `imagepush` part

mccutchen:
Fixed in #64

PR #66, opened 2021-10-13
adding method to return the hostname where httpbin is running

uromahn:
I wanted a method that identifies the host httpbin is running on and returns the hostname.


mccutchen:
Hey @uromahn, I'm sorry for the long delay in getting back to this.  Would you mind letting me know what your use case is for this feature?

Having thought about it for a while, I no longer think this is a good feature to add.  The rest of go-httpbin's functionality is about the incoming HTTP request, the client making the request, or exercising interesting characteristics in HTTP responses.

This would be the first feature to intentionally[1] reveal something about the underlying infrastructure, and I don't want to add that unless there's a good reason to do so, and I can't think of one myself!

<sub>_[1] I say "intentionally" here because some infra details are likely leaked in response headers added by the platform where a go-httpbin instance is hosted, but that's relatively unavoidable._</sub>

mccutchen:
Closing for now, will reopen if there's a compelling use case for this functionality!

x70b1:
I understand the argument that this would be the first infrastructure related information and maybe it fits not perfectly into httpbin.

My experience: I was debugging around on a haproxy and worked on the load balancing setup in a containerized environment.
I found httpbin and go-httpbin. It was the perfect tool as it gives a lot of information about the request details and so on. But I also asked me "On wich backend webserver are I am at the moment?". I found this PR but unfortunately it was closed.

A workaround was to add the requested server as information to the header, so that httpbin returns this value in its response back to my browser. But it would be much easier if httpbin could just return this. (Or something else where I can distinguish between 2 httpbin instances.) So I would like to see this feature.

mccutchen:
Ah, that is an interesting use case!  If we were to enable this functionality, I'd want to make it opt-in only to avoid unexpectedly revealing more information about existing go-httpbin instances than we were previously revealing.

Would that work for your use case?

x70b1:
Yeah, that's probably something that we don't want to expose by default.
Opt-in sounds good. A cli flag would be perfect.

mccutchen:
FYI for anyone watching this pull request: this was added in #81, but disabled by default.

PR #71, opened 2022-01-04
Fix link to Kenneth Reitz

st0le:


mccutchen:
Thanks @st0le!  Just needed to run `make` locally to regenerate the embedded static assets, but should be good to go now.  Will merge once we've got a ✅ from the tests.

PR #82, opened 2022-07-08
Run all tests in parallel

mccutchen:


httpbin/handlers_test.go
```diff
@@ -24,9 +24,11 @@ import (
 	"time"
 )
 
-const maxBodySize int64 = 1024 * 1024
-const maxDuration time.Duration = 1 * time.Second
-const alphanumLetters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+const (
+	maxBodySize     int64         = 1024
```
mccutchen, line 28: Dropping this from `1024 * 1024` to `1024` accounts for almost all of the speedup by reducing the worse-case `/drip` test case run in < 1s instead of 10s.

PR #84, opened 2022-10-12
Update golang to 1.19

x70b1:
I updated golang to 1.19.

Other than that I replaced `CMD` with `ENTRYPOINT`. This container image is a very good example for the difference between booth settings. You can now use the `CMD` for command line options. Example:

```
 nerdctl run mccutchen/go-httpbin -use-real-hostname
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,18 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.18
+# Golang
+FROM golang:1.19 as build
 
 WORKDIR /go/src/github.com/mccutchen/go-httpbin
 
 COPY . .
+
 RUN --mount=type=cache,id=gobuild,target=/root/.cache/go-build \
     make build buildtests
 
+# distroless
 FROM gcr.io/distroless/base
-COPY --from=0 /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
+
+COPY --from=build /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
+
 EXPOSE 8080
-CMD ["/bin/go-httpbin"]
+ENTRYPOINT ["/bin/go-httpbin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 # syntax = docker/dockerfile:1.3
-# Golang
-FROM golang:1.19 as build
+FROM golang:1.19 AS build
 
 WORKDIR /go/src/github.com/mccutchen/go-httpbin
 
@@ -9,10 +8,10 @@ COPY . .
 RUN --mount=type=cache,id=gobuild,target=/root/.cache/go-build \
     make build buildtests
 
-# distroless
 FROM gcr.io/distroless/base
 
 COPY --from=build /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
 
 EXPOSE 8080
+
 ENTRYPOINT ["/bin/go-httpbin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,5 +13,4 @@ FROM gcr.io/distroless/base
 COPY --from=build /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
 
 EXPOSE 8080
-
-ENTRYPOINT ["/bin/go-httpbin"]
+CMD ["/bin/go-httpbin"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,13 +1,18 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.18
+# Golang
+FROM golang:1.19 as build
```
mccutchen, line 3: Minor style preference:  Please capitalize `AS`

Dockerfile
```diff
@@ -1,13 +1,18 @@
...
 
+# distroless
 FROM gcr.io/distroless/base
-COPY --from=0 /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
+
+COPY --from=build /go/src/github.com/mccutchen/go-httpbin/dist/go-httpbin* /bin/
+
 EXPOSE 8080
-CMD ["/bin/go-httpbin"]
+ENTRYPOINT ["/bin/go-httpbin"]
```
mccutchen, line 18: I understand the advantage of this approach, _especially_ in a "distroless" image that doesn't include anything useful aside from the binary itself, but I'm afraid we need to keep the current behavior as-is because I know this image is used in a variety of CI processes and I'd rather not break them without good reason.

(I do wish I had built it this way from the start!)

Dockerfile
```diff
@@ -1,13 +1,18 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.18
+# Golang
```
mccutchen, line 2: This comment doesn't really add anything helpful, please drop it

Dockerfile
```diff
@@ -1,13 +1,18 @@
...
+FROM golang:1.19 as build
 
 WORKDIR /go/src/github.com/mccutchen/go-httpbin
 
 COPY . .
+
 RUN --mount=type=cache,id=gobuild,target=/root/.cache/go-build \
     make build buildtests
 
+# distroless
```
mccutchen, line 12: Likewise, please drop this as well

mccutchen (commented):
Thanks for the contribution!  I'm 👍 to bump to 1.19, but unfortunately I'm not comfortable making the `CMD` → `ENTRYPOINT` change for backwards-compatibility reasons.

To complete the version bump, please also update the versions in the GitHub Actions workflows:
- https://github.com/mccutchen/go-httpbin/blob/main/.github/workflows/test.yaml#L22
- https://github.com/mccutchen/go-httpbin/blob/main/.github/workflows/test.yaml#L41-L42 (we run regression tests against the previous two releases)
- https://github.com/mccutchen/go-httpbin/blob/main/.github/workflows/lint.yaml#L17

Dockerfile
x70b1: I see the point. This would be a break for all who use the docker image with a custom `CMD`. For a regular run it would not break something.

The question is: What would be a good reason? Ich would say this is a good one. It adds a common feature to thie image. And CI is there to fix issues as soon as they appear. 

So a release with a breaking change note would be useful?

mccutchen: I don't think the benefits of making this change outweigh the downsides.  To me, this feels like a weird case for a breaking change release; no actual functionality/API changes, just a change that might impact folks running the app via docker.

For now, I'll keep thinking about this.  As I said, I wish I had done it from the beginning, so I do see the advantages.  And I'd guess that _most_ people who use the docker image are configuring via env vars and would not notice the difference at all, but that's just a guess.

And the next time I do need to make an actual breaking change to the project's API/functionality or make a major version update I'll take the opportunity to make this change, as well.

mccutchen: … a goofy idea just occurred to me:  We could conceivably maintain backwards compatibility for existing users by updating `main()` to handle the case where it's called with `"go-httpbin"` at argv[1], so that both of these commands would work:

```
nerdctl run mccutchen/go-httpbin go-httpbin -use-real-hostname
```
```
nerdctl run mccutchen/go-httpbin -use-real-hostname
```

I can't yet tell whether that's a terrible idea, but something to consider.

x70b1: Sounds like a lot of work compared to the result. 
I think keeping this until yor create a bigger release is the best idea then.

mccutchen: Added an issue to help myself remember this https://github.com/mccutchen/go-httpbin/issues/85

mccutchen:
Ugh, looks like golangci-lint is not working w/ go 1.19.  Gonna push a commit or two up here to upgrade it, too, and see what happens.  We'll get this landed soon, thanks for your patience!

mccutchen:
🤦 or … I'll accidentally merge this when trying to push a fixup commit to this branch, how about that?

PR #86, opened 2022-10-17
Add test http client and example for doing a full ziti system

camotts:


mccutchen:
Intriguing!  While direct integration w/ openziti is probably never going to make sense for this project, if you end up putting up a `go-httpbin-ziti` fork/extension somewhere I'd be happy to link to it!

PR #89, opened 2022-11-10
Improve compatibility with httpbin for requests with bodies

anuraaga:
- Return for GET to /anything as httpbin does. While not common, it is possible to have bodies in GET requests with HTTP/1.0
- Don't escape HTML in JSON to reproduce httpbin behavior
- Always return a newline at end of JSON response

Great project, tests complete much faster than using python httpbin :) I am running [ftw](https://github.com/coreruleset/go-ftw), a test suite for WAFs with expectations based on httpbin. Many many tests including uncommon ones like HTTP/1.0 so hopefully it's helped squeeze out remaining incompatibilities. With this PR, the suite passes fine.

FYI, integrating with the code from this PR in https://github.com/corazawaf/coraza-proxy-wasm/pull/77

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -61,7 +61,7 @@ func (h *HTTPBin) Get(w http.ResponseWriter, r *http.Request) {
 // Anything returns anything that is passed to request.
 func (h *HTTPBin) Anything(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
-	case "GET", "HEAD":
+	case "HEAD":
 		h.Get(w, r)
 	default:
 		h.RequestWithBody(w, r)
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -54,7 +54,7 @@ func (h *HTTPBin) Get(w http.ResponseWriter, r *http.Request) {
 		Origin:  getClientIP(r),
 		URL:     getURL(r).String(),
 	}
-	body, _ := json.Marshal(resp)
+	body, _ := jsonMarshalNoEscape(resp)
 	writeJSON(w, body, http.StatusOK)
 }
 
@@ -83,7 +83,7 @@ func (h *HTTPBin) RequestWithBody(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	body, _ := json.Marshal(resp)
+	body, _ := jsonMarshalNoEscape(resp)
 	writeJSON(w, body, http.StatusOK)
 }
 
@@ -94,7 +94,7 @@ func (h *HTTPBin) Gzip(w http.ResponseWriter, r *http.Request) {
 		Origin:  getClientIP(r),
 		Gzipped: true,
 	}
-	body, _ := json.Marshal(resp)
+	body, _ := jsonMarshalNoEscape(resp)
 
 	buf := &bytes.Buffer{}
 	gzw := gzip.NewWriter(buf)
@@ -114,7 +114,7 @@ func (h *HTTPBin) Deflate(w http.ResponseWriter, r *http.Request) {
 		Origin:   getClientIP(r),
 		Deflated: true,
 	}
-	body, _ := json.Marshal(resp)
+	body, _ := jsonMarshalNoEscape(resp)
 
 	buf := &bytes.Buffer{}
 	w2 := zlib.NewWriter(buf)
@@ -129,7 +129,7 @@ func (h *HTTPBin) Deflate(w http.ResponseWriter, r *http.Request) {
 
 // IP echoes the IP address of the incoming request
 func (h *HTTPBin) IP(w http.ResponseWriter, r *http.Request) {
-	body, _ := json.Marshal(&ipResponse{
+	body, _ := jsonMarshalNoEscape(&ipResponse{
 		Origin: getClientIP(r),
 	})
 	writeJSON(w, body, http.StatusOK)
@@ -137,7 +137,7 @@ func (h *HTTPBin) IP(w http.ResponseWriter, r *http.Request) {
 
 // UserAgent echoes the incoming User-Agent header
 func (h *HTTPBin) UserAgent(w http.ResponseWriter, r *http.Request) {
-	body, _ := json.Marshal(&userAgentResponse{
+	body, _ := jsonMarshalNoEscape(&userAgentResponse{
 		UserAgent: r.Header.Get("User-Agent"),
 	})
 	writeJSON(w, body, http.StatusOK)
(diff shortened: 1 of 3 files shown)


httpbin/handlers.go
```diff
@@ -517,9 +517,8 @@ func (h *HTTPBin) Stream(w http.ResponseWriter, r *http.Request) {
 	f := w.(http.Flusher)
 	for i := 0; i < n; i++ {
 		resp.ID = i
-		line, _ := json.Marshal(resp)
+		line, _ := jsonMarshalNoEscape(resp)
 		w.Write(line)
-		w.Write([]byte("\n"))
```
anuraaga, line 522: This was needed because `json.Marshal` doesn't add a newline while `.Encode` does. I confirmed that httpbin seems to return a newline at the end of responses

httpbin/handlers.go
```diff
@@ -1016,8 +1015,18 @@ func (h *HTTPBin) Bearer(w http.ResponseWriter, r *http.Request) {
...
 	})
 	writeJSON(w, body, http.StatusOK)
 }
+
+// json.Marshal escapes HTML in strings while httpbin does not, so
+// we need to set up the encoder manually to reproduce that behavior.
+func jsonMarshalNoEscape(value interface{}) ([]byte, error) {
+	buffer := &bytes.Buffer{}
+	encoder := json.NewEncoder(buffer)
+	encoder.SetEscapeHTML(false)
```
mccutchen, line 1029: I'm not sure I understand what problem this is solving.  AFAICT, go-httpbin is already returning unescaped HTML in the response:

```python
>>> import requests
>>> requests.post('https://httpbingo.org/post', data="<html><body><h1>hello world</h1></body></html>", headers={"Content-Type": "text/html"}).json()["data"]
'<html><body><h1>hello world</h1></body></html>'
```

```bash
$ curl -sSL -X POST -d "<html><body><h1>hello world</h1></body></html>" -H "Content-Type: text/html" https://httpbingo.org/post | jq -r .data
<html><body><h1>hello world</h1></body></html>
```

Looking at the "raw" output from curl might shed a little light here, as it seems the `<` and `>` characters in the data are encoded as unicode references:

```
"data":"\u003chtml\u003e\u003cbody\u003e\u003ch1\u003ehello world\u003c/h1\u003e\u003c/body\u003e\u003c/html\u003e"
```

But, as you can see in the python and jq examples above, this form and the unencoded form are semantically identical.  What tool are you using that can't handle this form?

(Please let me know if I'm missing some nuance here!)

jcchavezs:
Ping @mccutchen

jcchavezs:
Related https://github.com/corazawaf/coraza-proxy-wasm/pull/77#issuecomment-1310144252

httpbin/handlers.go
mccutchen: Never mind, having thought about this for a few more minutes, I think I understand the issue:  Ideally, we should just send back the incoming data byte-for-byte, where the current approach does change it.

But … I'm not sure how to reconcile that with the fact that it must be encoded as valid JSON, so there is always going to be some input data that will be transformed in the output.  In the simplest case, that might just mean escaping double quotes in ASCII text, but it gets much more complex.  E.g. let's look at what happens when we post binary data (in this case a 1x1 png image).

Original httpbin returns it as a base64-encoded `data:` URL:

```
$ curl -sSL --data-binary @tiny.png -H "Content-Type: image/png"  https://httpbin.org/post | jq .data
"data:application/octet-stream;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAAAAAA6fptVAAAACklEQVQIW2O4AwAA3gDddw4iTwAAAABJRU5ErkJggg=="
```

go-httpbin returns it basically as-is, accounting for encoding the bytes in JSON-compatible format:

```
$ curl -sSL --data-binary @tiny.png -H "Content-Type: image/png"  https://httpbingo.org/post | jq .data
"�PNG\r\n\u001a\n\u0000\u0000\u0000\rIHDR\u0000\u0000\u0000\u0001\u0000\u0000\u0000\u0001\b\u0000\u0000\u0000\u0000:~�U\u0000\u0000\u0000\nIDAT\b[c�\u0003\u0000\u0000�\u0000�w\u000e\"O\u0000\u0000\u0000\u0000IEND�B`�"
```

So, basically, I'm not sure what the right call is here.

httpbin/helpers.go
```diff
@@ -123,12 +123,13 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
 	ct := r.Header.Get("Content-Type")
 	switch {
 	case strings.HasPrefix(ct, "application/x-www-form-urlencoded"):
-		// r.ParseForm() does not populate r.PostForm for DELETE requests, but
+		// r.ParseForm() does not populate r.PostForm for DELETE or GET requests, but
 		// we need it to for compatibility with the httpbin implementation, so
 		// we trick it with this ugly hack.
-		if r.Method == http.MethodDelete {
+		if r.Method == http.MethodDelete || r.Method == http.MethodGet {
+			method := r.Method
```
mccutchen, line 130: For clarity, let's call this `originalMethod`

mccutchen (commented):
Awesome, thanks for the contribution!

Overall, I'm sold on allowing GET requests with bodies in the `/anything` endpoint for compatibility's sake.

I'm a bit more conflicted on the JSON _encoding_ — **not** escaping, despite what the golang stdlib calls it — question, though.  I think it would help if you can provide a bit more detail about why the encoding matters.

httpbin/handlers.go
mccutchen: Okay, I'm back to being in favor of this change, after actually reading [the documentation for SetEscapeHTML](https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML):

> SetEscapeHTML specifies whether problematic HTML characters should be escaped inside JSON quoted strings. The default behavior is to escape &, <, and > to \u0026, \u003c, and \u003e to avoid certain safety problems that can arise when embedding JSON in HTML. 

I was initially worried that it would be doing something more invasive than this.  Seems like a reasonable update to me.

mccutchen:
@anuraaga to summarize my rambling above:  I'm 👍 on this change, just requesting a small fix to a variable name.

httpbin/handlers.go
anuraaga: Thanks - yup both forms are valid JSON, but unescaped allows the endpoint to be something like an echo server, which is how coreruleset tests are using it

anuraaga (commented):
Thanks! Fixed the variable

httpbin/handlers.go
mccutchen: > yup both forms are valid JSON, but unescaped allows the endpoint to be something like an echo server, which is how coreruleset tests are using it

I'm still fine with this change, but I'd suggest that if coreruleset is comparing JSON output byte-for-byte then it is far more brittle than it should be.

Will changes in the formatting of the JSON, like pretty-printing, break this in the future?

mccutchen:
Closing the loop here, these changes were [released in v2.5.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.5.0)

PR #94, opened 2022-11-11
Rework JSON serialization

mccutchen:
Instead of serializing JSON to a temporary slice of bytes before writing to the response, instead write directly to the response.

Note: With this change, we rely much more on the underlying net/http server's implicit handling of the Content-Length header. One notable side effect of this is that responses to HEAD requests no longer include a Content-Length.

@anuraaga's JSON serialization improvements in https://github.com/mccutchen/go-httpbin/pull/89 made me take a look at how we're handling JSON in general.  This change preserves those improvements, and adds pretty-printing for better readability (and to match the original httpbin).


httpbin/handlers.go
```diff
@@ -518,8 +512,8 @@ func (h *HTTPBin) Stream(w http.ResponseWriter, r *http.Request) {
 	f := w.(http.Flusher)
 	for i := 0; i < n; i++ {
 		resp.ID = i
-		line, _ := jsonMarshalNoEscape(resp)
-		w.Write(line)
+		line, _ := json.Marshal(resp)
```
anuraaga, line 515:
```suggestion
		// Call json.Marshal directly to avoid pretty printing
		line, _ := json.Marshal(resp)
```

anuraaga (approved):
Cool!

httpbin/handlers.go
mccutchen: Yep, good call, thanks!

PR #98, opened 2022-11-12
Refactor package main to enable testing of CLI interface

mccutchen:
I may come to regret the brittleness and busywork this might add to this project's test suite, but: I didn't like how much untested argument/env parsing logic lived inside the main function, so I factored it out into a testable package.

One thing missing from _fully_ testing the app is the ability to actually run it and test graceful server shutdown.  Maybe later.


mccutchen:
Overall test coverage has dropped, but this adds test coverage to a previously untested bit of complex logic.

PR #107, opened 2023-01-15
feat: ensure that 404 responses indicate that they originate from go-httpbin

chinaran:
When go-httpbin is used as a test service and accessed through a gateway, 404 may be returned due to incorrect configuration of the gateway. 
By optimizing the response content, you can clearly know whether the 404 is returned by the gateway or go-httpbin.

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -23,7 +23,8 @@ func notImplementedHandler(w http.ResponseWriter, r *http.Request) {
 // Index renders an HTML index page
 func (h *HTTPBin) Index(w http.ResponseWriter, r *http.Request) {
 	if r.URL.Path != "/" {
-		http.Error(w, "Not Found", http.StatusNotFound)
+		msg := fmt.Sprintf("Not Found (go-httpbin does not have the API: %s)", r.URL.Path)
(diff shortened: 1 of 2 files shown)


httpbin/handlers.go
```diff
@@ -23,7 +23,8 @@ func notImplementedHandler(w http.ResponseWriter, r *http.Request) {
 // Index renders an HTML index page
 func (h *HTTPBin) Index(w http.ResponseWriter, r *http.Request) {
 	if r.URL.Path != "/" {
-		http.Error(w, "Not Found", http.StatusNotFound)
+		msg := fmt.Sprintf("Not Found (go-httpbin does not have the API: %s)", r.URL.Path)
```
mccutchen, line 26: How about 

> Not Found (go-httpbin does not handle the path %s)

as a slight tweak to the message? I just don’t think of these paths/routes as “APIs”.

mccutchen (approved):
👍 I like this idea, thanks! 

Left one minor suggestion for the message below, but LGTM in general!

httpbin/handlers.go
chinaran: I agree with you, thanks, have adjusted the code.

PR #111, opened 2023-02-14
upgrade go to 1.20

rarkebauer:
Upgrade Go version to 1.20

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
-          go-version: '1.19'
-      - uses: golangci/golangci-lint-action@v3.2.0
+          go-version: '1.20'
+      - uses: golangci/golangci-lint-action@v3.4.0
         with:
-          version: v1.50.0
+          version: v1.51.1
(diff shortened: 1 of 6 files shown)


mccutchen:
Nice, thanks!  Looks like linting is failing due to 1.20's deprecation of insecure use of the non-cryptographic rng:

```
  Error: SA1019: rand.Read has been deprecated since Go 1.20 because it shouldn't be used: For almost all use cases, crypto/rand.Read is more appropriate.  (staticcheck)
  Error: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access.  (staticcheck)
  Error: SA1019: rand.Read has been deprecated since Go 1.20 because it shouldn't be used: For almost all use cases, crypto/rand.Read is more appropriate.  (staticcheck)
```

Do you want to take a pass a fixing these issues, or should I?  I'm hoping they'll be trivial to update, but I haven't taken a look yet myself!

rarkebauer:
I'm happy to go ahead and take a look at fixing those issues! Thanks for the quick review

mccutchen (approved):
👍 thanks for the contribution!

PR #117, opened 2023-04-24
Fix #90 Return binary request bodies as base64 string

### LINKED ISSUE #90: Match original httpbin's handling of binary data
As discovered in https://github.com/mccutchen/go-httpbin/pull/89#discussion_r1019231548, the original httpbin implementation takes an interesting approach to handling binary data that we might want to match.  Copying from that comment:

Let's look at what happens when we post binary data (in this case a 1x1 png image).

Original httpbin returns it as a base64-encoded `data:` URL:

```
$ curl -sSL --data-binary @tiny.png -H "Content-Type: image/png"  https://httpbin.org/post | jq .data
"data:application/octet-stream;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAAAAAA6fptVAAAACklEQVQIW2O4AwAA3gDddw4iTwAAAABJRU5ErkJggg=="
```

go-httpbin returns it basically as-is, accounting for encoding the bytes in JSON-compatible format:

```
$ curl -sSL --data-binary @tiny.png -H "Content-Type: image/png"  https://httpbingo.org/post | jq .data
"�PNG\r\n\u001a\n\u0000\u0000\u0000\rIHDR\u0000\u0000\u0000\u0001\u0000\u0000\u0000\u0001\b\u0000\u0000\u0000\u0000:~�U\u0000\u0000\u0000\nIDAT\b[c�\u0003\u0000\u0000�\u0000�w\u000e\"O\u0000\u0000\u0000\u0000IEND�B`�"
```

gab1one:
I implemented the binary encoding like it is done in the original httpbin #90.
I am not sure if the covered `Content-Type` are sufficient, but the list should be easy to extend.

### DIFF
--- a/httpbin/handlers_test.go
+++ b/httpbin/handlers_test.go
@@ -6,6 +6,7 @@ import (
 	"compress/gzip"
 	"compress/zlib"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -544,6 +545,7 @@ func testRequestWithBody(t *testing.T, verb, path string) {
 		testRequestWithBodyMultiPartBody,
 		testRequestWithBodyQueryParams,
 		testRequestWithBodyQueryParamsAndBody,
+		testRequestWithBodyBinaryBody,
 	}
 	for _, testFunc := range testFuncs {
 		testFunc := testFunc
@@ -551,6 +553,54 @@ func testRequestWithBody(t *testing.T, verb, path string) {
 		t.Run(getTestName(verb, testFunc), func(t *testing.T) {
 			t.Parallel()
 			testFunc(t, verb, path)
+		})
+	}
+}
+
+func testRequestWithBodyBinaryBody(t *testing.T, verb string, path string) {
+	tests := []struct {
+		contentType  string
+		bodyResponse string
+	}{
+		{"application/octet-stream", "encodeMe"},
+		{"image/png", "encodeMe-png"},
+		{"image/webp", "encodeMe-webp"},
+		{"image/jpeg", "encodeMe-jpeg"},
+	}
+	for _, test := range tests {
+		test := test
+		t.Run("content type/"+test.contentType, func(t *testing.T) {
+			t.Parallel()
+
+			testBody := bytes.NewReader([]byte(test.bodyResponse))
+
+			r, _ := http.NewRequest(verb, path, testBody)
+			r.Header.Set("Content-Type", test.contentType)
+			w := httptest.NewRecorder()
+			app.ServeHTTP(w, r)
+
+			assertStatusCode(t, w, http.StatusOK)
+			assertContentType(t, w, jsonContentType)
+
+			var resp *bodyResponse
+			err := json.Unmarshal(w.Body.Bytes(), &resp)
+			if err != nil {
+				t.Fatalf("failed to unmarshal body %s from JSON: %s", w.Body, err)
+			}
+
+			if resp.Data != "data:application/octet-stream;base64,"+base64.StdEncoding.EncodeToString([]byte(test.bodyResponse)) {
+				t.Fatalf("expected binary encoded response data, got %#v", resp.Data)
+			}
+			if resp.JSON != nil {
+				t.Fatalf("expected nil response json, got %#v", resp.JSON)
+			}
+
+			if len(resp.Args) > 0 {
+				t.Fatalf("expected no query params, got %#v", resp.Args)
+			}
+			if len(resp.Form) > 0 {
+				t.Fatalf("expected no form data, got %#v", resp.Form)
+			}
 		})
 	}
 }
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -159,6 +159,19 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
 		if err != nil && err != io.EOF {
 			return err
 		}
+	// in case of binary data, we encode it as base64 and return it as a data url
+	case strings.HasPrefix(ct, "image/png"):
+		fallthrough
+	case strings.HasPrefix(ct, "image/jpeg"):
+		fallthrough
+	case strings.HasPrefix(ct, "image/webp"):
+		fallthrough
+	case strings.HasPrefix(ct, "application/octet-stream"):
+		data := base64.StdEncoding.EncodeToString(body)
+		if err != nil && err != io.EOF {
+			return err
+		}
+		resp.Data = string("data:application/octet-stream;base64," + data)
 	}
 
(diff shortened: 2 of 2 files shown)


httpbin/helpers.go
```diff
@@ -159,6 +159,19 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
...
+	case strings.HasPrefix(ct, "image/jpeg"):
+		fallthrough
+	case strings.HasPrefix(ct, "image/webp"):
+		fallthrough
+	case strings.HasPrefix(ct, "application/octet-stream"):
+		data := base64.StdEncoding.EncodeToString(body)
+		if err != nil && err != io.EOF {
+			return err
+		}
+		resp.Data = string("data:application/octet-stream;base64," + data)
```
mccutchen, line 174: It would be nice to preserve the incoming content type in the output (though httpbin.org does not do this, AFAICT)

httpbin/helpers.go
```diff
@@ -159,6 +159,19 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
...
 			return err
 		}
+	// in case of binary data, we encode it as base64 and return it as a data url
+	case strings.HasPrefix(ct, "image/png"):
+		fallthrough
+	case strings.HasPrefix(ct, "image/jpeg"):
+		fallthrough
+	case strings.HasPrefix(ct, "image/webp"):
+		fallthrough
+	case strings.HasPrefix(ct, "application/octet-stream"):
```
mccutchen, line 169: I think this is a reasonable starting point, but I also got curious about httpbin's implementation.

It turns out [they do something fairly naive](https://github.com/postmanlabs/httpbin/blob/f8ec666b4d1b654e4ff6aedd356f510dcac09f83/httpbin/helpers.py#L85-L106):  If the incoming request body either a) cannot be decoded from utf8 or b) cannot be re-serialized to JSON, it is treated as binary data.

I wonder if we should do something similar here, instead of working with this fixed list of "probably binary" content-type headers?

mccutchen: Maybe we should do both:  Keep this set of "probably binary" headers for which we _know_ we want return a base64-encoded `data:` url, but add a case to handle the situation where we've got unknown non-utf8 data in the incoming request body?

httpbin/handlers_test.go
```diff
@@ -555,6 +557,54 @@ func testRequestWithBody(t *testing.T, verb, path string) {
 	}
 }
 
+func testRequestWithBodyBinaryBody(t *testing.T, verb string, path string) {
+	tests := []struct {
+		contentType  string
+		bodyResponse string
```
mccutchen, line 563: Naming nit:  I think this should be called `requestBody`, since it's really defining the data we're sending in the request body.

httpbin/helpers.go
gab1one: Good idea, so that we return something like `data:image/png;base64`  when supplied with an image ?

httpbin/helpers.go
gab1one: I pushed a new commit that handles all non-text content this way.

httpbin/handlers_test.go
gab1one: fixed

gab1one:
@mccutchen are there anything missing, or something else you want me to add ?

httpbin/helpers.go
```diff
@@ -159,11 +166,29 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
...
+		resp.Data = encodeData(body, "image/png")
+	case strings.HasPrefix(ct, "image/jpeg"):
+		resp.Data = encodeData(body, "image/jpeg")
+	case strings.HasPrefix(ct, "image/webp"):
+		resp.Data = encodeData(body, "image/webp")
+	case strings.HasPrefix(ct, "application/octet-stream"):
+		resp.Data = encodeData(body, "application/octet-stream")
+	default:
+		// If we don't know how to handle the content type, we'll just return it encoded as base64 data url
+		resp.Data = encodeData(body, ct)
```
mccutchen, line 180: Are there any downsides to simplifying all of this down to something like so?

```go
// for any other content-type, encode incoming request body as base64
// and return it as a data url
default:
    resp.Data = encodeData(body, ct)
```

I _think_ we'd need to handle content types like `text/html; charset=utf-8`, where we'd only want to echo back the part before the first `;` (if any).  I suppose that could be done directly in `encodeData()`, or it could be done before this switch statement starts (which would also potentially simplify each of the other cases by eliminating the need for `strings.HasPrefix` for most of them).

httpbin/helpers.go
```diff
@@ -133,6 +133,13 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
 
 	ct := r.Header.Get("Content-Type")
 	switch {
+	// cases where we don't need to parse the body
+	case ct == "":
+		fallthrough
+	case strings.HasPrefix(ct, "text/"):
```
mccutchen, line 139: Let's do the same for `"html/"`?

httpbin/helpers.go
```diff
@@ -133,6 +133,13 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
 
 	ct := r.Header.Get("Content-Type")
 	switch {
+	// cases where we don't need to parse the body
+	case ct == "":
+		fallthrough
```
mccutchen, line 138: I think `ct == ""` should be treated as if `ct == "application/octet-stream"` (i.e., if no content type is given, just treat the body as opaque bytes).

mccutchen (commented):
I'm sorry for the delayed reply.  This is looking good, but I've got a couple more suggestions.  Let me know what you think!

httpbin/helpers.go
gab1one: I had it set up like that first, but then a lot of tests failed, that resulted in this test failing: https://github.com/mccutchen/go-httpbin/blob/main/httpbin/handlers_test.go#L659

httpbin/helpers.go
gab1one: good point, that makes the code much clearer. I'll push a commit that does just that.

httpbin/helpers.go
gab1one: I pushed a second commit, that implements this change, I adapted that one test.

mccutchen (approved):
This came out really nice, thanks for the contribution (and the patience with the review process) @gab1one!

mccutchen:
Nice! This was released as v2.7.0 yesterday, in case you missed it:
https://github.com/mccutchen/go-httpbin/releases/tag/v2.7.0

PR #120, opened 2023-05-19
feat: support serving under a path prefix

waschik:
Make relative links in main help page. Some links (like `absolute-redirekt` and `links`) will not work with prefix. But they still work without.

Could be tested with this program:
```go
package main

import (
	"github.com/mccutchen/go-httpbin/v2/httpbin"
	"net/http"
)

const (
	prefix = "/prefix"
)

func main() {
	var handler http.Handler = httpbin.New()
	handler = http.StripPrefix(prefix, handler)

	mux := http.NewServeMux()
	mux.Handle(prefix+"/", handler)
	http.ListenAndServe(":8090", mux)
}
```

While running this result can be checked at http://localhost:8090/prefix/


mccutchen:
Ah, interesting use case, this isn't something I've considered before.  Unfortunately, fixing the links on the index page is an incomplete solution; as you mentioned, any endpoint that generates links to other endpoints (like `/redirect`, `/absolute-redirect`, etc) will break:

```
$ curl -v http://localhost:8090/prefix/redirect/5 2>&1 | grep -e '> GET' -e '< HTTP' -e '< Location'
> GET /prefix/redirect/5 HTTP/1.1
< HTTP/1.1 302 Found
< Location: /relative-redirect/4
```

So, if we want go-httpbin to properly function when being served under a path prefix, we need a more holistic solution.  Probably adding an optional `--prefix` configuration that would be used to compute URLs.  (Unfortunately, this would also require us to turn the static index.html into a template of some sort, but we can probably keep that pretty simple and render it only once at startup and continue serving it quickly from memory.)

Are you interested in taking on that challenge?

waschik:
Okay. I tried to add prefix to code and make the index html and form post a template. Executable size increased on my system by about 20% (7.5 MB to 8.9 MB). Because software then needs to know the prefix I made an option like you suggested and also called the `StripPrefix` inside handler creation.

Then after
```bash
./go-httpbin -port 8090 -prefix /prefix &
```
I could run the following:
```
$ curl -v http://localhost:8090/prefix/redirect/5 2>&1 | grep -e '> GET' -e '< HTTP' -e '< Location'
> GET /prefix/redirect/5 HTTP/1.1
< HTTP/1.1 302 Found
< Location: /prefix/relative-redirect/4
```

I did know change all unit tests. This would make tests more complicated (mostly for-loops with and without prefix). I do not know what you prefer. Another option (maybe not the nicest) would be that I write a complete wrapper and do some search and replacement in result body and headers (e.g. `Location` header).

mloskot:
@mccutchen 
> Ah, interesting use case, this isn't something I've considered before.

I sometimes deploy go-httpbin  behind Kong proxy, for testing, and where I apply path-based routing, so the service is available at e.g. https://example.com/tests/httpbin and it would be nice if the endpoints on the frontpage could be generated to

/tests/httpbin/anything
/tests/httpbin/base64
...

mccutchen:
Yeah, my apologies to @waschik and anyone else interested in this change.  I dropped the ball on getting this reviewed and merged, and _then_ I refactored a few things along with the whole dang test suite, leading to a bunch of churn that will need to be addressed.

I'll try to make time this weekend circle back to it!

waschik:
@mccutchen I tried to fixed the merge failures. Hope you did not do them parallel and I home I do not have to do it again.

Do you know https://github.com/stretchr/testify? Looks a bit similar to your assert. With that I would not have to add a assert.Contains method which I needed because of two tests of body.

httpbin/handlers_test.go
```diff
@@ -2897,10 +2935,10 @@ func TestHostname(t *testing.T) {
 
 		realHostname := "real-hostname"
 		app := New(WithHostname(realHostname))
-		srv, client := newTestServer(app)
-		defer srv.Close()
+		env := newTestEnvironment(app)
+		defer env.srv.Close()
 
-		req, err := http.NewRequest("GET", srv.URL+"/hostname", nil)
+		req, err := http.NewRequest("GET", env.srv.URL+"/hostname", nil)
```
mccutchen, line 2941: I don't think these changes are needed?

httpbin/handlers.go
```diff
@@ -178,31 +178,31 @@ var (
...
 <title>Multiple Choices</title>
 </head>
 <body>
 <ul>
-<li><a href="/image/jpeg">/image/jpeg</a></li>
-<li><a href="/image/png">/image/png</a></li>
-<li><a href="/image/svg">/image/svg</a></li>
+<li><a href="%[1]s/image/jpeg">/image/jpeg</a></li>
+<li><a href="%[1]s/image/png">/image/png</a></li>
+<li><a href="%[1]s/image/svg">/image/svg</a></li>
```
mccutchen, line 189: Wow, TIL about this `%[1]s` syntax, that's very handy!

httpbin/handlers.go
```diff
@@ -464,7 +464,7 @@ func (h *HTTPBin) DeleteCookies(w http.ResponseWriter, r *http.Request) {
 			Expires:  time.Now().Add(-1 * 24 * 365 * time.Hour),
 		})
 	}
-	w.Header().Set("Location", "/cookies")
+	w.Header().Set("Location", h.prefix+"/cookies")
 	w.WriteHeader(http.StatusFound)
```
mccutchen, line 468: Given how often this pattern now shows up across handlers, I propose that we wrap the logic up in a helper method:

```golang
func (h *HTTPBin) doRedirect(path string, code int) {
    w.Header().Set("Location", h.prefix+"/cookies")
    w.WriteHeader(code)
}
```

(That name would clash with the _existing_ `doRedirect()` helper, which is a bit higher level and serves only the `Redirect`/`AbsoluteRedirect`/`RelativeRedirect` handler.  I'd suggest we rename the existing `doRedirect` to something like `handleRedirect` instead.)

httpbin/handlers_test.go
```diff
@@ -33,21 +33,28 @@ import (
...
 	maxBodySize int64         = 1024
 	maxDuration time.Duration = 1 * time.Second
+	testPrefix                = "/a-prefix"
 )
 
-// "Global" test app, server, & client to be reused across test cases.
-// Initialized in TestMain.
-var (
+type environment struct {
 	app    *HTTPBin
```
mccutchen, line 40: Seems like we don't actually need this `app` field

cmd/go-httpbin/.gitignore
```diff
@@ -0,0 +1 @@
+/go-httpbin
```
mccutchen, line 1: Not sure why this is necessary, but I'd prefer all ignores to be consolidated in the top-level `.gitignore` file.

If you use `make`/`make run`/`make watch`/`make test` during development, this probably wouldn't be needed!

mccutchen (commented):
Thanks a lot for picking this back up!  It's looking good overall, but I've left a bit of feedback below.  Please let me know if it makes sense to you!

One more high level comment:  Would you mind adding a `.tmpl` extension to the end of the two files that have change from static html to golang templates?  I.e. `index.html` -> `index.html.tmpl`.

httpbin/handlers_test.go
mccutchen: Ah, maybe they're necessary because `newTestServer` was deleted?  I guess we could keep it around, and make `newTestEnvironment` use it?

mccutchen:
> Do you know https://github.com/stretchr/testify? Looks a bit similar to your assert. With that I would not have to add a assert.Contains method which I needed because of two tests of body.

Yes, I use that and a variety of other testing helper packages in other projects.  For this project, an explicit goal is using zero dependencies outside the stdlib, in order to make it easier for _other_ projects to decide to pull it into their own test suites.

httpbin/handlers.go
waschik: You mean a `doRedirectCookies` method or `cookies` a parameter to some method?

mccutchen: I mean there are now a bunch of places where we've got code like this, where we're constructing a redirect URL and must remember to include the prefix, so I'd like to have a helper method like the one I showed above.

_This_ code would then become something like

```go
h.doRedirect("/cookies", http.StatusFound)
return
```

and the new `doRedirect()` helper would take care of handling the prefix.

Where it gets messy is that there's already an **existing**, higher level helper method `doRedirect()` helper that will need to be renamed.  I think renaming the existing helper to `handleRedirect()` makes sense, as that better captures what the existing helper is actually doing.

httpbin/handlers_test.go
waschik: You are right. I removed it. Was originally needed when there was no test server.

httpbin/httpbin.go
```diff
@@ -55,6 +55,13 @@ type HTTPBin struct {
 	// The app's http handler
 	handler http.Handler
 
+	prefix string
+
+	index_html      []byte
+	forms_post_html []byte
```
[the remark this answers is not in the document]
waschik: Fixed it.

httpbin/static/index.html
```diff
@@ -58,65 +58,65 @@ <h1>go-httpbin(1)</h1>
...
-<li><a href="/user-agent"><code>/user-agent</code></a> Returns user-agent.</li>
-<li><a href="/uuid"><code>/uuid</code></a> Generates a <a href="https://en.wikipedia.org/wiki/Universally_unique_identifier">UUIDv4</a> value.</li>
-<li><a href="/websocket/echo?max_fragment_size=2048&amp;max_message_size=10240"><code>/websocket/echo?max_fragment_size=2048&amp;max_message_size=10240</code></a> A WebSocket echo service.</li>
-<li><a href="/xml"><code>/xml</code></a> Returns some XML</li>
+<li><a href="."><code>{{.Prefix}}/</code></a> This page.</li>
+<li><a href="absolute-redirect/6"><code>{{.Prefix}}/absolute-redirect/:n</code></a> 302 Absolute redirects <em>n</em> times.</li>
+<li><a href="anything"><code>{{.Prefix}}/anything/:anything</code></a> Returns anything that is passed to request.</li>
+<li><a href="base64/aHR0cGJpbmdvLm9yZw=="><code>{{.Prefix}}/base64/:value</code></a> Decodes a Base64-encoded string.</li>
+<li><a href="base64/decode/aHR0cGJpbmdvLm9yZw=="><code>{{.Prefix}}/base64/decode/:value</code></a> Explicit URL for decoding a Base64 encoded string.</li>
```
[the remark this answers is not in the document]
waschik: I fixed it. But I would prefer relative links. That was my original goal for the MR. All other stuff was only for completeness of redirect/location handlers. Thats why I named the branch `relative-doc-links`. I did not change documentation because that might change something. For href the browser will go to direct link.

httpbin/handlers_test.go
```diff
@@ -61,38 +68,58 @@ func TestMain(m *testing.M) {
...
+	defer env.srv.Close()
+
 	os.Exit(m.Run())
 }
 
 func TestIndex(t *testing.T) {
-	t.Run("ok", func(t *testing.T) {
-		t.Parallel()
+	for _, env := range envs {
+		t.Run("ok"+env.prefix, func(t *testing.T) {
```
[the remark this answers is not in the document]
waschik: Thanks. Sorry for forgetting. I'll use in future golint.

httpbin/handlers_test.go
waschik: Actually I replaced newTestServer with newTestEnvironment. Because there were only one call remaining which might use direct variables I changed it. But I changed it back.

httpbin/handlers.go
waschik: I renamed the `doRedirect` to `handleRedirect` and used at two places. Is there a reason why https://github.com/waschik/go-httpbin/blob/7e81a9365c6fe9f1a4691742792dcd2e3d8a1fa6/httpbin/handlers.go#L883 is using `r.URL.String()` and not `r.URL.Path`? If `r.URL.Path` would be okay it would be possible to replace it also with the new `doRedirect`. For `r.URL.String()` I am not sure if it will not sometimes include the schema and then prefix might be wrong.

For the absolute redirects new `doRedirect` would also not possible otherwise I get locations like `/a-prefixhttp://...`

cmd/go-httpbin/.gitignore
waschik: I moved it to toplevel gitignore. I added it so I do not risk to add it with `git add .` if if I manually build with `go build` inside the `cmd/go-httpbin` directory. But if you prefer I remove it. Sure it is not needed for the prefix.

httpbin/httpbin.go
```diff
@@ -55,6 +55,13 @@ type HTTPBin struct {
...
+
+	specialCases map[int]*statusCase
```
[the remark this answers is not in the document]
waschik: Maybe again `statusSpeciaCases` like in current state of your main branch? https://github.com/mccutchen/go-httpbin/blob/b63a792fd6a8d9289a9b7076defc1b58cbb50402/httpbin/handlers.go#L201-L280

The 40x status codes are not really redirects. So `redirectSpecialCases` might be (in my point of view) a bit confusing.

httpbin/handlers_test.go
```diff
@@ -61,38 +68,58 @@ func TestMain(m *testing.M) {
...
+	var env *environment
+	app = createApp()
+	env = newTestEnvironment(app)
+	defer env.srv.Close()
+	srv = env.srv
+	client = env.client
+	envs = append(envs, env)
+	defaultEnv = env
+
+	env = newTestEnvironment(createApp(WithPrefix(testPrefix)))
```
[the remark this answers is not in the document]
waschik: Sorry for that. Was too much change. I fixed this. Sure the `append` was intended.

httpbin/handlers.go
waschik: :smile: Would be possible to use this also use this instead of the templates if that dependency should be avoided. But I do not think this would make the code cleaner and all `%` would have to be escaped.

httpbin/static/index.html.tmpl
waschik, line 1: > Thanks a lot for picking this back up! It's looking good overall, but I've left a bit of feedback below. Please let me know if it makes sense to you!
> 
> One more high level comment: Would you mind adding a `.tmpl` extension to the end of the two files that have change from static html to golang templates? I.e. `index.html` -> `index.html.tmpl`.

I changed it. But I had no good experience with template suffixes (e.g. `.j2`). I sometimes loose editor support like syntax highlighting if I use normal extension. In the case case editors are also not confused by the `{{ }}`: Quotes are still matching.

httpbin/static/index.html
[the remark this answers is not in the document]
mccutchen: I'm curious, why do you prefer relative links here?  I think it should accomplish the exact same thing, and I just prefer the explicitness of absolute links (even though they're more verbose in the template).

httpbin/handlers.go
mccutchen: > Is there a reason why https://github.com/waschik/go-httpbin/blob/7e81a9365c6fe9f1a4691742792dcd2e3d8a1fa6/httpbin/handlers.go#L883 is using `r.URL.String()` and not `r.URL.Path`?

I _think_ this is to ensure that any query params are carried through the redirects, but it looks like we don't have any explicit test cases covering that behavior.  Let's keep it as-is for now.

> For the absolute redirects new `doRedirect` would also not possible otherwise I get locations like `/a-prefixhttp://...`

Ah that's a good point.  Maybe he new `doRedirect()` should change its behavior based on whether the given argument starts with `http://` or `https://`?

httpbin/cmd/cmd.go
```diff
@@ -142,6 +146,7 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 	fs.IntVar(&cfg.ListenPort, "port", defaultListenPort, "Port to listen on")
 	fs.StringVar(&cfg.rawAllowedRedirectDomains, "allowed-redirect-domains", "", "Comma-separated list of domains the /redirect-to endpoint will allow")
 	fs.StringVar(&cfg.ListenHost, "host", defaultListenHost, "Host to listen on")
+	fs.StringVar(&cfg.Prefix, "prefix", "", "Path prefix (empty or start with slash and does not end with slash)")
```
mccutchen, line 149: Every config option may be set by CLI arguments _and_ by environment variables.  I think we need to add support for setting the prefix via a `PREFIX` env var.

We also need to update this section of the README to note the new option:
https://github.com/mccutchen/go-httpbin/blob/main/README.md#configuration

mccutchen (commented):
Almost there!  Thanks a lot for the effort you've put into this @waschik, I know it turned into a much bigger change than you originally intended.

httpbin/handlers.go
waschik: Maybe it works if prefix is only added if arguments start with `/`. I'll experiment a bit with this. Otherwise option would be to parse URL or add another parameter.

mccutchen: Yep, I think pivoting on whether the first character is `/` would be fine here!

httpbin/httpbin.go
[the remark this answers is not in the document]
waschik: I renamed it back.

httpbin/static/index.html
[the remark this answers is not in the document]
waschik: For relative links it would be possible to set prefix also from outside (e.g. with nginx) without setting prefix if you do not care for the redirects. So I think relative will not hurt. But sure as it is it works now.

httpbin/handlers.go
waschik: I implemented this with slash checking. Wrong adding of prefix of external prefix was shown in test cases. Hope it is correct now.

mccutchen (approved):
👍 LGTM!  Really nice work here @waschik, thanks again for your persistence in getting this implemented!

mccutchen:
FYI @waschik @mloskot these changes were released in [v2.13.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.13.0)

mloskot:
Thank you @waschik & @mccutchen !

mloskot:
The prefix works like a charm!

----

image

----

I just had to tweak Ingress for Kong Ingress Controller in my cluster to 1) match path used for path-based with the prefix 2) not not strip the path when passing request to the go-httpbin backend, for example:

```yaml
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: httpbin
  namespace: tests
  annotations:
    konghq.com/strip-path: "false"
spec:
  ingressClassName: kong
  rules:
  - host: my.example.com
    http:
      paths:
      - path: /tests/httpbin
        pathType: ImplementationSpecific
        backend:
          service:
            name: httpbin
            port:
              number: 8080
```

PR #122, opened 2023-06-14
Add websocket echo endpoint

wrhight:
Adds a handler at "/wscho" that accepts websocket upgrades and echoes any messages written to it. 

Inspired by the example set in https://github.com/nhooyr/websocket/tree/master/examples/echo

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,8 @@
 module github.com/mccutchen/go-httpbin/v2
 
 go 1.16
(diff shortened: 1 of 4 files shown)


wrhight:
mistakes were made. didn't mean to open this upstream - my bad!

mccutchen:
No worries, hope you’re working on something interesting!

Issue #91, opened 2022-11-10
Tracking compatibility with original httpbin


mccutchen:
Let's use this as a place to track compatibility with original httpbin:

- [ ] https://github.com/mccutchen/go-httpbin/issues/15
- [ ] https://github.com/mccutchen/go-httpbin/issues/37
- [ ] https://github.com/mccutchen/go-httpbin/pull/89
- [ ] https://github.com/mccutchen/go-httpbin/issues/90

sonbui00:
I got a different response 

```
Original
 curl https://httpbin.org/response-headers\?access_token\=faketoken\&token_type\=Bearer   
{
  "Content-Length": "128", 
  "Content-Type": "application/json", 
  "access_token": "faketoken", 
  "token_type": "Bearer"
}

go-httpbin
 curl http://httpbin:9100/response-headers\?access_token\=faketoken\&token_type\=Bearer  
{
  "access_token": [
    "faketoken"
  ],
  "token_type": [
    "Bearer"
  ]
}
```

mccutchen:
Good catch, thanks. I think we should include both Content-Length and Content-Type to match original httpbin.

Are you interested in opening an issue for that specific feature and/or opening a pull request to implement it? (An interesting question is what should happen if the request specifies different, incorrect values for those two headers; might be worth seeing how httpbin.org handles those cases.)

But, to be clear, go-httpbin will continue to serialize header values (and form values and any other “zero or more” values inherent to HTTP) as JSON arrays for consistency, where original httpbin serializes them as single scalar values, arrays, or comma-separated strings, depending on the endpoint and incoming request parameters.

bleggett:
it should be noted that `httpbin` returns the following body for 418

```
        -=[ teapot ]=-

           _...._
         .'  _ _ `.
        | ."` ^ `". _,
        \_;`"---"`|//
          |       ;/
          \_     _/
            `"""`
```

and this implementation does not.

As per https://www.rfc-editor.org/rfc/rfc2324#section-2.3.2 this is strictly speaking in compliance with the RFC, but it is a difference in behavior.

mccutchen:
@bleggett please report that as a separate issue (or skip the separate issue and open a pull request to fix it).

PR #123, opened 2023-06-18
feat: add method property to responses

llimllib:
```console
# before: httpbingo lacks the "method" field but httpbin.org has it
$ curl -s https://httpbingo.org/anything | jq .method                                                                                                        9:31PM
null
$ curl -s https://httpbin.org/anything | jq .method
"GET"

# after: httpbingo has the method field!
$ curl -s http://0.0.0.0:8080/anything | jq .method
"GET"
```

This was mentioned in #6, but then not listed in #91

- I took the highly technical testing approach of adding a test for `Method` anywhere in `handlers_test.go` that mentioned `Args`; do you desire more tests? fewer tests?
- Anywhere else I should add this field that I didn't already?

For what it's worth, I discovered this when working on upgrading the testing for our [httpsnippet library](https://github.com/readmeio/httpsnippet); I wanted to use `go-httpbin` but our tests expect the method field to be present in the responses

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -47,6 +47,7 @@ func (h *HTTPBin) Get(w http.ResponseWriter, r *http.Request) {
 	writeJSON(http.StatusOK, w, &noBodyResponse{
 		Args:    r.URL.Query(),
 		Headers: getRequestHeaders(r),
+		Method:  r.Method,
 		Origin:  getClientIP(r),
 		URL:     getURL(r).String(),
 	})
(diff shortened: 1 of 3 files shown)


mccutchen (approved):
Wow, if someone had asked me whether go-httpin returns the request method as part of the response body here, I think I would have confidently said yes.  Oops!

Thanks a lot for the contribution, LGTM pending a CI test run.

mccutchen:
FYI, this was released in [v2.9.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.9.0)

PR #125, opened 2023-06-18
fix: don't update Data when body is empty

### LINKED ISSUE #124: `data` property should be empty on a GET to `/anything`
It's up to you to decide of course, but I think this should be considered a bug:

```
$ curl -s 'https://httpbingo.org/anything?one=two' | jq .data
"data:application/octet-stream;base64,"
```

The `data` field for a `GET` request, as I understand it, should always be empty.

update: now I looked at the tests and I see you're emulating httpbin, which will handle GETs with bodies. Which is weird! But if the request _doesn't_ have a body, it still should have an empty `data` field.

(Not that you need to match httpbin byte for byte, but it does return the empty string here)

llimllib:
When a request to `/anything` has an empty body, the `data` field of the output should be the empty string. Currently, `go-httpbin is returning `data:application/octet-stream;base64,`.

- Should the output actually be null? That would match `files`, `form`, and `json`, but also would require deeper changes since `Data` is stored as a string not a pointer, and I didn't want to mess with things any more than I had to. An empty string works fine for my purposes
- how can I test this PR? I tried but I got confused by the TestAnything method
- all the current tests pass under this change

This PR closes #124. Before:

```
$ curl -s 'http://0.0.0.0:8080/anything?one=two'
{
  "args": {
    "one": [
      "two"
    ]
  },
  "headers": {
    "Accept": [
      "*/*"
    ],
    "Host": [
      "0.0.0.0:8080"
    ],
    "User-Agent": [
      "curl/7.88.1"
    ]
  },
  "method": "GET",
  "origin": "127.0.0.1:60402",
  "url": "http://0.0.0.0:8080/anything?one=two",
  "data": "data:application/octet-stream;base64,",
  "files": null,
  "form": null,
  "json": null
}
```

After:
```
$ curl -s 'http://0.0.0.0:8080/anything?one=two'
{
  "args": {
    "one": [
      "two"
    ]
  },
  "headers": {
    "Accept": [
      "*/*"
    ],
    "Host": [
      "0.0.0.0:8080"
    ],
    "User-Agent": [
      "curl/7.88.1"
    ]
  },
  "method": "GET",
  "origin": "127.0.0.1:60595",
  "url": "http://0.0.0.0:8080/anything?one=two",
  "data": "",
  "files": null,
  "form": null,
  "json": null
}
```

### DIFF
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -126,6 +126,11 @@ func parseBody(w http.ResponseWriter, r *http.Request, resp *bodyResponse) error
 	}
 	resp.Data = string(body)
 
+	// if we read an empty body, there's no need to do anything further
+	if len(resp.Data) == 0 {
+		return nil
+	}
+
 	// After reading the body to populate resp.Data, we need to re-wrap it in
 	// an io.Reader for further processing below
 	r.Body.Close()


mccutchen:
+1, I agree that this change makes sense.

> * Should the output actually be null? That would match `files`, `form`, and `json`, but also would require deeper changes since `Data` is stored as a string not a pointer, and I didn't want to mess with things any more than I had to. An empty string works fine for my purposes

That's a good question, let me think about it a little bit this morning.  My first instinct is that an empty string is a reasonable enough compromise that, even if `null` would be more consistent with other optional elements of these responses, it's probably not worth the effort it would take to coerce the golang type system & json serializer into handling this for us.

I might take a quick pass locally just to see how feasible something like a `type OptionalData string` w/ custom marshal/unmarshal methods might be, but I also feel like every time I try to do this a) it takes me forever to figure out and b) it doesn't end up working quite the way I want it to.

> * how can I test this PR? I tried but I got confused by the TestAnything method

Also a good question, and I'm sorry for any confusion!  I know the test suite is a bit of a mess, and I have no idea off the top of my head how best to test this, but lemme take a look.

mccutchen:
> That's a good question, let me think about it a little bit this morning. My first instinct is that an empty string is a reasonable enough compromise that, even if `null` would be more consistent with other optional elements of these responses, it's probably not worth the effort it would take to coerce the golang type system & json serializer into handling this for us.

Okay, took a quick pass at this, and I don't think it's worth all of the additional overhead that would be required to marshal empty bodies as `null` in the response.

The custom type approach feels like a whole lot of complexity (including a particularly heavyweight custom `MarshalJSON` method that would have to allocate an extra temporary buffer in order to use a json.Encoder in order to disable HTML escaping in order to maintain compatibility with [this other compatibility-related change](https://github.com/mccutchen/go-httpbin/pull/89)), and for now I don't particularly want to get into the business of working with string pointers everywhere.

mccutchen:
> > * how can I test this PR? I tried but I got confused by the TestAnything method
> 
> Also a good question, and I'm sorry for any confusion! I know the test suite is a bit of a mess, and I have no idea off the top of my head how best to test this, but lemme take a look.

Oh boy, I just spent wayyyy too long trying to figure out why the existing test cases for requests with empty bodies were showing a different behavior from the curl test (either the tests should have failed, or the curl should have returned an empty string for the `"data"` field).

Turns out it's down to a couple of things:
- Most/all of the tests in the test suite use `httptest.NewRecorder` and just call the ServeHTTP method directly, rather than spinning up a live HTTP server using `httptest.NewServer` and making an actual HTTP request, like so: https://github.com/mccutchen/go-httpbin/blob/1223a7364acfd61265fe6d3e74f3b58518d63064/httpbin/handlers_test.go#L633-L636
- After adding some logging, it became clear that this conditional is true when using the test recorder and false when making a real HTTP request: https://github.com/mccutchen/go-httpbin/blob/1223a7364acfd61265fe6d3e74f3b58518d63064/httpbin/helpers.go#L115-L118

So, weirdly, we already have unit tests in place that describe the behavior we're actually _implementing_ in this pull request, they just weren't testing the actual live behavior in the face of an actual HTTP request.

My inclination is to land this change as-is and then I'll follow up with an update to the test suite to stop using the recorder approach.  I've considered that a number of times in the past for vague notions of correctness, but this is the first time I can remember running into an actual behavior difference that mattered quite so much.

mccutchen:
FYI, this was released in [v2.9.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.9.0)

llimllib:
Thanks so much! I want to add that I've found this project really useful, thanks for all the work

PR #126, opened 2023-06-21
Return https if r.TLS is not nil

llimllib:
`go-httpbin` was incorrectly returning `http` as the scheme for its URL if you ran it with TLS specified via `HTTPS_CERT_FILE` and `HTTPS_KEY_FILE`. Update it to return https as the scheme in this case by checking if `r.TLS` is not nil

Before:

```
$ mkcert test
$ docker run -e HTTPS_CERT_FILE='/tmp/test.pem' -e HTTPS_KEY_FILE='/tmp/test-key.pem' -p 8080:8080 -v $(pwd):/tmp mccutchen/go-httpbin
$ curl -sk https://localhost:8080/get | jq -r .url
http://localhost:8080/get
```

After (TLS):

```
$ docker run -e HTTPS_CERT_FILE='/tmp/test.pem' -e HTTPS_KEY_FILE='/tmp/test-key.pem' -p 8080:8080 -v $(pwd):/tmp llimllib/go-httpbin
$ curl -sk https://localhost:8080/get | jq -r .url
https://localhost:8080/get
```

After (no TLS):

```
$ docker run -p 8080:8080 llimllib/go-httpbin
$ curl -sk http://localhost:8080/get | jq -r .url
http://localhost:8080/get
```

I got the idea for how to check the scheme from [this SO post](https://stackoverflow.com/a/61446922/42559)

**testing**: added a test for `getURL`

### DIFF
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -59,6 +59,9 @@ func getURL(r *http.Request) *url.URL {
 	if scheme == "" && r.Header.Get("X-Forwarded-Ssl") == "on" {
 		scheme = "https"
 	}
+	if scheme == "" && r.TLS != nil {
+		scheme = "https"
+	}
 	if scheme == "" {
 		scheme = "http"
 	}
--- a/httpbin/helpers_test.go
+++ b/httpbin/helpers_test.go
@@ -1,9 +1,11 @@
 package httpbin
 
 import (
+	"crypto/tls"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


mccutchen:
> **testing**: I'm not sure how to test this in some way that isn't completely tautological, open to ideas

Heh, FWIW I'd have no objection to a fairly contrived (or tautological) test for `getURL()`.  We've already got something similar for `getClientIP()`:
https://github.com/mccutchen/go-httpbin/blob/7e8d543c0f1ca838b5840cae45662b44a7aff025/httpbin/helpers_test.go#L167-L217

llimllib:
OK, thanks for the nudge I added a pretty thorough test for `getURL`

mccutchen (approved):
Nice, thanks!

Issue #15, opened 2019-03-05
[compat] original httpbin combines mulitple header values into comma-separated string


johnroach:
Hi, started using your httpbin thank you for porting this work to Go!

I have been having a problem when using this solution for headers.

I am wondering if you would be OK with simply returning all headers as a string like httpbin instead of setting as list types. It looks like that is the default behavior for httpbin.org

```
$ curl -H 'foo: bar' -H 'foo: baz' https://httpbin.org/get
{
  "args": {},
  "headers": {
    "Accept": "*/*",
    "Foo": "bar,baz",        <===========================
    "Host": "httpbin.org",
    "User-Agent": "curl/7.54.0"
  },
  ...
  "url": "https://httpbin.org/get"
}
```

vs 

```
$ curl -H 'foo: bar' -H 'foo: baz' https://httpbin.mccutchen.awesome.com/get
{
   "url" : "https://httpbin.mccutchen.awesome.com/get",
   "origin" : "100.112.202.128",
   "args" : {},
   "headers" : {
      "Foo" : [ "bar", "baz" ],     <===============================

   }
}
```

mccutchen:
The behavior of the original httpbin service used to be that it would return a single string for single-valued headers or a list of strings for multi-valued headers, and I just adopted the latter form for consistency and ease of implementation.  I'm not sure when httpbin switched to returning multiple headers as a single, comma-separated string.

For the moment, I'm 👎 on making this change, because the current behavior seems more "correct" to me.  It is an unfortunate break in compatibility with upstream httpbin, though.

Would you mind elaborating on the specific problem you're having with this approach?  I want to think about this for a bit, if that's okay!

(And sorry for the delayed reply, I was on vacation for a bit!)

johnroach:
No problem about the late reply. I hope you had an awesome vacation!

The issue is that our developers wrote a couple hundred tests against httpbin and I wanted to switch to using yours. However, this incompatibility breaks said tests. That is all really.

If you feel strongly about this I understand. That being said I really hope I can use your solution.

Have a great day!

mccutchen:
I've taken a pass at this and #37 from time to time over the last couple of years, but I'm never satisfied with the complexity that it requires.

Ultimately, I don't think this (admittedly large) part of backwards-compatibility is worth pursuing, and prefer the consistency of our existing approach:  any key-value pair where each key may correspond to one or more values will always be represented as a JSON array (AKA list, slice, etc) with one or more items.

[...]

PR #139, opened 2023-07-21
feat: allow filtering incoming request headers

bytemain:
I want to deploy go-httpbin to Alibaba Cloud Function Compute (A Serverless Runtime), But in FC, the Server will add some custom header on the request headers:

for example:
CleanShot 2023-07-21 at 12 50 06@2x

So I add this logic to process unused headers:

```golang
	opts := []httpbin.OptionFunc{
		httpbin.WithMaxBodySize(cfg.MaxBodySize),
		httpbin.WithMaxDuration(cfg.MaxDuration),
		httpbin.WithObserver(httpbin.StdLogObserver(logger)),
		httpbin.WithExcludeHeaders("x-ignore-*,x-info-this-key"),
	}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ variables (or a combination of the two):
 | `-max-duration` | `MAX_DURATION` | Maximum duration a response may take | 10s |
 | `-port` | `PORT` | Port to listen on | 8080 |
 | `-use-real-hostname` | `USE_REAL_HOSTNAME` | Expose real hostname as reported by os.Hostname() in the /hostname endpoint | false |
+| `-exclude-headers` | `EXCLUDE_HEADERS` | Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard matching. For example: `"foo,bar,x-fc-*"` | - |
 
 **Notes:**
 - Command line arguments take precedence over environment variables.
--- a/README.md
+++ b/README.md
@@ -160,6 +160,13 @@ public internet, consider tuning it appropriately:
      logging using [zerolog] and further hardens the HTTP server against
      malicious clients by tuning lower-level timeouts and limits.
 
+5. **Prevent leaking sensitive headers**
+
+  By default, go-httpbin will return any headers sent by the client in the response.
+  But if you want to deploy go-httpbin in some serverless environment, you may want to drop some headers.
+  You can use the `-exclude-headers` CLI argument or the `EXCLUDE_HEADERS` env var to configure an appropriate allowlist.
+  For example, Alibaba Cloud Function Compute will [add some headers like `x-fc-*` to the request](https://www.alibabacloud.com/help/en/fc/user-guide/specification-details). if you want to drop these `x-fc-*` headers, you can set `EXCLUDE_HEADERS=x-fc-*`.
+
 ## Development
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ variables (or a combination of the two):
 | `-max-duration` | `MAX_DURATION` | Maximum duration a response may take | 10s |
 | `-port` | `PORT` | Port to listen on | 8080 |
 | `-use-real-hostname` | `USE_REAL_HOSTNAME` | Expose real hostname as reported by os.Hostname() in the /hostname endpoint | false |
-| `-exclude-headers` | `EXCLUDE_HEADERS` | Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard matching. For example: `"foo,bar,x-fc-*"` | - |
+| `-exclude-headers` | `EXCLUDE_HEADERS` | Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard suffix matching. For example: `"foo,bar,x-fc-*"` | - |
 
 **Notes:**
 - Command line arguments take precedence over environment variables.
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -13,6 +13,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -21,6 +22,8 @@ import (
 
 // Base64MaxLen - Maximum input length for Base64 functions
 const Base64MaxLen = 2000
+
+type HeaderInterceptor func(h http.Header) http.Header
 
 // requestHeaders takes in incoming request and returns an http.Header map
 // suitable for inclusion in our response data structures.
@@ -435,4 +438,25 @@ func (b *base64Helper) Encode() ([]byte, error) {
 // Decode - decode data from base64
 func (b *base64Helper) Decode() ([]byte, error) {
 	return base64.URLEncoding.DecodeString(b.data)
+}
+
+func wildCardToRegexp(pattern string) string {
+	components := strings.Split(pattern, "*")
+	if len(components) == 1 {
+		// if len is 1, there are no *'s, return exact match pattern
+		return "^" + pattern + "$"
+	}
+	var result strings.Builder
+	for i, literal := range components {
+
+		// Replace * with .*
+		if i > 0 {
+			result.WriteString(".*")
+		}
+
+		// Quote any regular expression meta characters in the
+		// literal text.
+		result.WriteString(regexp.QuoteMeta(literal))
+	}
+	return "^" + result.String() + "$"
 }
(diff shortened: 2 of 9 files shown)


mccutchen:
Ah, interesting idea.  Given that using this new Interceptor will require wiring up your own go-httpbin server rather than just running a prebuilt binary or docker image, I should point out that the same thing could be accomplished via plain old stlib HTTP handler middleware without requiring any changes to go-httpbin.  (I'd be happy to show a specific example, if it would help!)

It seems like the real value would be exposing this as configuration, to make it easy for anyone to deploy a "stock" go-httpbin image and still drop platform-specific headers.  I can imagine being able to do something like this (including wildcard prefix matches) being generally useful:

```
-exclude-headers="Foo,Bar,X-Fc-*"
EXCLUDE_HEADERS="Foo,Bar,X-Fc-*"
```

Do you have any interest in adapting this work to that end?  (If so, I'd consider simplifying the implementation to just deal with headers.  The notion of flexible "Interceptors" is nice, but feels like a bit more than we need here.)

bytemain:
Yes, It seem that expose this feature as a configuration is more reasonable.

But in my opinion, I think the Interceptor mechanism should be kept, it make the whole httpbin more configurable. It is actually a hook system,  bring some hooks so that we can configure by ourself.

In the first version PR, I just added a very simple interceptor implementation(I am lazy, XD), I think the Observer is the same as Interceptor, So can I still use the "interceptor"?

mccutchen:
What other interceptors do you imagine being useful here?  Maybe it's a lack of imagination on my part, but I do not think we'd want to support anything other than modifying _incoming_ request headers, and I can't think of any reason to do anything aside from dropping them.

To do that, a fancy system of interceptor hooks is not required.  It can be implemented as a fairly simple middleware.

(I'd put the Observer in a different category, simply because golang's stdlib makes it difficult/awkward to instrument outgoing HTTP _responses_.  There, the Observer abstraction makes it possible for go-httpbin to surface details about the generated response in a way that is easy for 3rd parties to build on.  No such abstraction is necessary for filtering out incoming request headers.)

httpbin/httpbin.go
```diff
@@ -179,3 +185,38 @@ func (h *HTTPBin) Handler() http.Handler {
...
+}
+
+func (h *HTTPBin) SetExcludeHeaders(excludeHeaders string) {
+	// comma separated list of headers to exclude from response
+	tmp := strings.Split(excludeHeaders, ",")
+	result := make([]*regexp.Regexp, len(tmp))
+
+	for i, v := range tmp {
+		pattern := wildCardToRegexp(strings.TrimSpace(strings.ToLower(v)))
+		compiled, _ := regexp.Compile(pattern)
```
bytemain, line 217: Should we throw error if compiled regex failed?

bytemain:
@mccutchen I have simplified the implementation and add configuration ability~

httpbin/httpbin.go
```diff
@@ -179,3 +185,38 @@ func (h *HTTPBin) Handler() http.Handler {
...
+		if skipped {
+			continue
+		}
+		result[k] = v
+	}
+
+	return result
+}
+
+func (h *HTTPBin) SetExcludeHeaders(excludeHeaders string) {
```
mccutchen, line 210: This doesn't need to be part of the public API.  Let's inline this into the `WithExcludeHeaders` option func below (sort of like we inline the similar logic in `WithAllowedRedirects`).

httpbin/helpers.go
```diff
@@ -22,18 +23,23 @@ import (
 // Base64MaxLen - Maximum input length for Base64 functions
 const Base64MaxLen = 2000
 
+type HeaderInterceptor func(h http.Header) http.Header
```
mccutchen, line 26: This doesn't need to be part of the public API

httpbin/httpbin.go
```diff
@@ -179,3 +185,38 @@ func (h *HTTPBin) Handler() http.Handler {
...
+	for k, v := range headers {
+		lowerCase := strings.ToLower(k)
+		skipped := false
+		for _, pattern := range h.excludeHeadersPatterns {
+			matched := pattern.Match([]byte(lowerCase))
+			if matched {
+				skipped = true
+				break
+			}
+		}
```
mccutchen, line 200: I have a few ideas for optimizations here:
- Make this function return early if `h.excludeHeadersPatterns == nil`, which will be true in the vast majority of instances
- Build a single pattern that matches _all_ of excluded headers, so we only need to match each header key against a single pattern rather than this nested for loop
- Make sure the pattern matches case-insensitively so we do not need the separate `strings.ToLower()`

httpbin/httpbin.go
```diff
@@ -179,3 +185,38 @@ func (h *HTTPBin) Handler() http.Handler {
 
 	return handler
 }
+
+func (h *HTTPBin) HeadersProcessor(headers http.Header) http.Header {
```
mccutchen, line 189: This shouldn't be part of the public API, let's un-export it.

mccutchen (commented):
Nice, thanks!  I've got some feedback below, please let me know if it makes sense to you.  In particular, I think we can make the regex matching more efficient.

httpbin/httpbin.go
mccutchen: In this case, I'd be in favor of just using `regexp.MustCompile` and panicking on error.  I think it's fine for the process to fail loudly and obviously on initial startup if misconfigured.

bytemain:
Thanks for your advice, I have done for them.

And now the headerProcessor will only be used if user configured `exclude-headers` configuration.

README.md
```diff
@@ -91,6 +91,7 @@ variables (or a combination of the two):
 | `-max-duration` | `MAX_DURATION` | Maximum duration a response may take | 10s |
 | `-port` | `PORT` | Port to listen on | 8080 |
 | `-use-real-hostname` | `USE_REAL_HOSTNAME` | Expose real hostname as reported by os.Hostname() in the /hostname endpoint | false |
+| `-exclude-headers` | `EXCLUDE_HEADERS` | Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard matching. For example: `"foo,bar,x-fc-*"` | - |
```
mccutchen, line 94: Thanks for thinking to update the README!  Would you mind also adding a note under the **Production considerations** section mentioning this new config option as a way to prevent leaking sensitive headers?

httpbin/helpers.go
```diff
@@ -433,3 +437,62 @@ func (b *base64Helper) Encode() ([]byte, error) {
...
+		if i > 0 {
+			result.WriteString(".*")
+		}
+
+		// Quote any regular expression meta characters in the
+		// literal text.
+		result.WriteString(regexp.QuoteMeta(literal))
+	}
+	return "^" + result.String() + "$"
+}
```
mccutchen, line 460: I think we should simplify this a bit by only allowing a single trailing `*` as a wildcard.

README.md
mccutchen: Also, per my suggestion below about only allowing a single suffix wildcard, I might suggest a tweak here:

```suggestion
| `-exclude-headers` | `EXCLUDE_HEADERS` | Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard suffix matching. For example: `"foo,bar,x-fc-*"` | - |
```

mccutchen (commented):
This is looking great, thanks so much for iterating on it with me!  We're very close, I just have a couple of final requests below.

Plus, there are a few very fiddly test cases that assert the CLI's help output, which are failing due to the new CLI argument here:
https://github.com/mccutchen/go-httpbin/actions/runs/5664402459/job/15362579323?pr=139

The easiest way to fix those failures is to copy the `got:` value directly into the test case.  (I know this is very annoying, I'm sorry!)

httpbin/helpers.go
bytemain: `wildCardToRegexp` is a very simple logic, if we want only allowing trailing `*`, we need add extra code.

in my opinion, there is no need to change the behavior

mccutchen: My feeling is that the entire function would be much simpler if it only handled a single trailing wildcard, and that a single trailing wildcard will cover almost every real-world usecase. 

If you feel strongly about being able to match multiple wildcards, please add test coverage for
- a single wildcard in the middle of the string
- multiple wildcards in the string

bytemain: Yes, I have add some testcase about this func.

mccutchen (approved):
👍 LGTM, thanks a lot for the contribution @bytemain!

mccutchen:
Released in [v2.11.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.11.0).  Thanks again!

bytemain:
By the way, if we use the intercept mechanism, the developer can change this behavior:

- https://github.com/mccutchen/go-httpbin/issues/15

mccutchen:
… they could also do that in a simple middleware, by combining the header values into a comma-separated string before they reach go-httpbin. 

That middleware would not be much more complex than a custom interceptor.

Issue #141, opened 2023-08-07
/response-headers endpoint compatibility 


sonbui00:
I got a different response.
```
Original
 curl https://httpbin.org/response-headers\?access_token\=faketoken\&token_type\=Bearer   
{
  "Content-Length": "128", 
  "Content-Type": "application/json", 
  "access_token": "faketoken", 
  "token_type": "Bearer"
}

go-httpbin
 curl http://httpbin:9100/response-headers\?access_token\=faketoken\&token_type\=Bearer  
{
  "access_token": [
    "faketoken"
  ],
  "token_type": [
    "Bearer"
  ]
}
```

mccutchen:
I still think this is worth fixing, at least in part.  Copying and lightly adapting [a comment](https://github.com/mccutchen/go-httpbin/pull/142#issuecomment-1668668594) I left on the reporter's [abandoned attempt](https://github.com/mccutchen/go-httpbin/pull/142) to fix this issue:

---

If we're aiming for compatibility, it's worth exploring what the original httpbin's behavior is in weird circumstances? Like, what happens with a request like `/response-headers?Content-Length=999999999`? I think we need some test coverage for at least that edge case, to codify whatever behavior we decide on.

Turns out, the behavior is interesting (TLS-related noise elided below):

```
$ curl -v --http1.1 'https://httpbin.org/response-headers?Content-Type=text/plain&Content-Length=99999'
*   Trying 54.210.149.139:443...
* Connected to httpbin.org (54.210.149.139) port 443 (#0)
> GET /response-headers?Content-Type=text/plain&Content-Length=99999 HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Mon, 07 Aug 2023 22:40:24 GMT
< Content-Type: text/plain
< Content-Length: 99999
< Connection: keep-alive
< Server: gunicorn/19.9.0
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
< 
{
  "Content-Length": [
    "122", 
    "99999"
  ], 
  "Content-Type": [
    "application/json", 
    "text/plain"
  ]
}
* transfer closed with 99877 bytes remaining to read
* Closing connection 0
curl: (18) transfer closed with 99877 bytes remaining to read
```

[...]

Issue #37, opened 2020-08-11
[compat] original httpbin returns either lists or scalars depending on input data


splitbrain:
The answers between httpbin and httpbingo differ. httpbingo will always return an array for each parameter or header, while httpbin will return simple objects and use sub arrays only when needed. I think your behaviour is maybe a more sensible approach, but it's harder to read for humans and makes it hard to use httpbingo as a drop-in replacement for httpbin.

```
curl -s  'http://httpbin.org/get?foo=bar&foo=baz&what=this' |jq
{
  "args": {
    "foo": [
      "bar",
      "baz"
    ],
    "what": "this"
  },
...
```

```
$ curl -s  'http://httpbingo.org/get?foo=bar&foo=baz&what=this' |jq 
{
  "args": {
    "foo": [
      "bar",
      "baz"
    ],
    "what": [
      "this"
    ]
  },
...
```

splitbrain:
I just noticed that this is somewhat a duplicate of #15 but not quite.

mccutchen:
Yeah, I'm really torn on this.  While making httpbingo.org a drop-in replacement for (at least the most common/basic functionality of) httpbin.org would be great, that particular pattern is particularly ugly to implement in Go.  Maybe I'll take another pass at this and see how bad it really is.

waitingsong:
I think common behaviour is:
```
GET  /get?foo=12&bar=23
```
Should got 
```json
{ 
 foo: 12,
 bar: 23,
}
```

and then
```
GET  /get?foo[]=12&foo[]=23
```
Should got 
```json
{ 
 foo: [12, 23],
}
```

mccutchen:
> and then
> 
> ```
> GET  /get?foo[]=12&foo[]=23
> ```
> 
> Should got
> 
> ```json
> { 
>  foo: ["12", "23"],
> }
> ```

Minor note:  There is no special handling of `?foo[]=12&foo[]=23` style of query param in either go-httpbin or the original httpbin.  That request will result in a response like

```json
{
    "foo[]": ["12", "23"]
}
```

in both projects.  Here's original httpbin:

```
$ curl -s 'https://httpbin.org/get?foo[]=1&foo[]=2' | jq .args
{
  "foo[]": [
    "1",
    "2"
  ]
}
```

and go-httbin:

```
$ curl -s 'https://httpbingo.org/get?foo[]=1&foo[]=2' | jq .args
{
  "foo[]": [
    "1",
    "2"
  ]
}
[...]

PR #142, opened 2023-08-07
fix: API response-headers compatibility with origin #131

sonbui00:
API response-headers compatibility with the origin
Fix: https://github.com/mccutchen/go-httpbin/issues/141 

```
curl  "http://127.0.0.1:8080/response-headers?Bar=bar1&Bar=bar2&Foo=foo"                   
{
  "Bar": [
    "bar1",
    "bar2"
  ],
  "Content-Length": 104,
  "Content-Type": "application/json; charset=utf-8",
  "Foo": "foo"
}

curl "https://httpbin.org/response-headers?Bar=bar1&Bar=bar2&Foo=foo"
{
  "Bar": [
    "bar1", 
    "bar2"
  ], 
  "Content-Length": "127", 
  "Content-Type": "application/json", 
  "Foo": "foo"
}
```

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -314,15 +314,26 @@ func (h *HTTPBin) Unstable(w http.ResponseWriter, r *http.Request) {
 // ResponseHeaders responds with a map of header values
 func (h *HTTPBin) ResponseHeaders(w http.ResponseWriter, r *http.Request) {
 	args := r.URL.Query()
+	body := make(map[string]interface{})
 	for k, vs := range args {
 		for _, v := range vs {
 			w.Header().Add(k, v)
+		}
+		if len(vs) == 1 {
+			body[k] = vs[0]
+		} else {
+			body[k] = vs
 		}
 	}
 	if contentType := w.Header().Get("Content-Type"); contentType == "" {
 		w.Header().Set("Content-Type", jsonContentType)
 	}
-	mustMarshalJSON(w, args)
+	body["Content-Type"] = w.Header().Get("Content-Type")
+	body["Content-Length"] = 0
+	buf := &bytes.Buffer{}
+	json.NewEncoder(buf).Encode(body)
+	body["Content-Length"] = buf.Len()
+	mustMarshalJSON(w, body)
 }
 
 func redirectLocation(r *http.Request, relative bool, n int) string {
--- a/httpbin/handlers_test.go
+++ b/httpbin/handlers_test.go
@@ -1163,7 +1163,18 @@ func TestResponseHeaders(t *testing.T) {
 
 		req, _ := http.NewRequest("GET", fmt.Sprintf("%s/response-headers?%s", srv.URL, wantHeaders.Encode()), nil)
 		resp := must.DoReq(t, client, req)
-		result := mustParseResponse[http.Header](t, resp)
+
+		type bodyResponse struct {
+			Foo           string   `json:"Foo"`
+			Bar           []string `json:"Bar"`
+			ContentType   string   `json:"Content-Type"`
+			ContentLength int      `json:"Content-Length"`
+		}
+
+		result := mustParseResponse[bodyResponse](t, resp)
+
+		assert.ContentType(t, resp, result.ContentType)
+		assert.Equal[bool](t, result.ContentLength > 0, true, "JSON response Content-Length mismatch")
 
 		for k, expectedValues := range wantHeaders {
 			// expected headers should be present in the HTTP response itself
@@ -1171,8 +1182,13 @@ func TestResponseHeaders(t *testing.T) {
 			assert.DeepEqual(t, respValues, expectedValues, "HTTP response headers mismatch")
 
 			// they should also be reflected in the decoded JSON resposne
-			resultValues := result[k]
-			assert.DeepEqual(t, resultValues, expectedValues, "JSON response headers mismatch")
+			if len(expectedValues) == 1 {
+				value := reflect.ValueOf(result).FieldByName(k).String()
+				assert.Equal[string](t, value, expectedValues[0], "JSON response headers mismatch")
+			} else {
+				value := reflect.ValueOf(result).FieldByName(k).Interface().([]string)
+				assert.DeepEqual(t, value, expectedValues, "JSON response headers mismatch")
+			}
 		}
 	})
 


httpbin/handlers.go
```diff
@@ -314,15 +314,26 @@ func (h *HTTPBin) Unstable(w http.ResponseWriter, r *http.Request) {
...
+	body := make(map[string]interface{})
 	for k, vs := range args {
 		for _, v := range vs {
 			w.Header().Add(k, v)
 		}
+		if len(vs) == 1 {
+			body[k] = vs[0]
+		} else {
+			body[k] = vs
+		}
```
mccutchen, line 326: Please drop this logic.  Per [my comment here](https://github.com/mccutchen/go-httpbin/issues/91#issuecomment-1667003449)

> But, to be clear, go-httpbin will continue to serialize header values (and form values and any other “zero or more” values inherent to HTTP) as JSON arrays for consistency, where original httpbin serializes them as single scalar values, arrays, or comma-separated strings, depending on the endpoint and incoming request parameters.

That might not be very clear.  To restate it, what I mean is that go-httpbin returns all header values as JSON arrays, even if only one header is given.

I know that is not actually backwards-compatible with the original httpbin, but for now I've decided that I prefer the consistency (and implementation simplicity) of this approach and I'm okay with the incompatibility.  (See https://github.com/mccutchen/go-httpbin/issues/37 and https://github.com/mccutchen/go-httpbin/issues/15 for related discussion.)

httpbin/handlers.go
```diff
@@ -314,15 +314,26 @@ func (h *HTTPBin) Unstable(w http.ResponseWriter, r *http.Request) {
 // ResponseHeaders responds with a map of header values
 func (h *HTTPBin) ResponseHeaders(w http.ResponseWriter, r *http.Request) {
 	args := r.URL.Query()
+	body := make(map[string]interface{})
```
mccutchen, line 317: With my comment below, I don't think we need anything more than `args` here.

httpbin/handlers.go
```diff
@@ -314,15 +314,26 @@ func (h *HTTPBin) Unstable(w http.ResponseWriter, r *http.Request) {
...
 	if contentType := w.Header().Get("Content-Type"); contentType == "" {
 		w.Header().Set("Content-Type", jsonContentType)
 	}
-	mustMarshalJSON(w, args)
+	body["Content-Type"] = w.Header().Get("Content-Type")
+	body["Content-Length"] = 0
+	buf := &bytes.Buffer{}
+	json.NewEncoder(buf).Encode(body)
+	body["Content-Length"] = buf.Len()
+	mustMarshalJSON(w, body)
```
mccutchen, line 336: This is a bit of a strange approach, double-encoding the body just to get the content length.  Please rewrite it to do the encoding once.  The existing [`writeResponse()` helper func](https://github.com/mccutchen/go-httpbin/blob/62ec5553f2401137c24a0cc67f04c5fd278a4e4b/httpbin/helpers.go#L96-L100) might be useful here.

httpbin/handlers_test.go
```diff
@@ -1163,16 +1163,32 @@ func TestResponseHeaders(t *testing.T) {
...
+			Foo           string   `json:"Foo"`
+			Bar           []string `json:"Bar"`
+			ContentType   string   `json:"Content-Type"`
+			ContentLength int      `json:"Content-Length"`
+		}
+
+		result := mustParseResponse[bodyResponse](t, resp)
+
+		assert.ContentType(t, resp, result.ContentType)
+		assert.Equal[bool](t, result.ContentLength > 0, true, "JSON response Content-Length mismatch")
```
mccutchen, line 1177: I think we can do away with most of this specialization if we simplify the response as requested above.

mccutchen (commented):
Thanks for the contribution!  I've left a few comments below that will help simplify these changes, please let me know if they make sense to you.

Also, have you explored what the original httpbin's behavior is in weird circumstances?  Like, what happens with a request like `/response-headers?Content-Length=999999999`?  I think we need some test coverage for at least that edge case, to codify whatever behavior we decide on.

mccutchen:
(Also, I apologize in advance for slow replies on my part, I am traveling right now.)

mccutchen:
> Also, have you explored what the original httpbin's behavior is in weird circumstances? Like, what happens with a request like `/response-headers?Content-Length=999999999`? I think we need some test coverage for at least that edge case, to codify whatever behavior we decide on.

The behavior is interesting (TLS-related noise elided below):

```
$ curl -v --http1.1 'https://httpbin.org/response-headers?Content-Type=text/plain&Content-Length=99999'
*   Trying 54.210.149.139:443...
* Connected to httpbin.org (54.210.149.139) port 443 (#0)
> GET /response-headers?Content-Type=text/plain&Content-Length=99999 HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Mon, 07 Aug 2023 22:40:24 GMT
< Content-Type: text/plain
< Content-Length: 99999
< Connection: keep-alive
< Server: gunicorn/19.9.0
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
< 
{
  "Content-Length": [
    "122", 
    "99999"
  ], 
  "Content-Type": [
    "application/json", 
    "text/plain"
  ]
}
* transfer closed with 99877 bytes remaining to read
* Closing connection 0
curl: (18) transfer closed with 99877 bytes remaining to read
```

I can convince myself that letting the incoming request to this specific endpoint dictate the actual Content-Length of the response could be useful for testing weird behavior in HTTP client or proxy code, so I'd be in favor of matching the Python implementation there.

mccutchen:
> How can I get just key - value?

I'm sorry, but I don't understand what you mean.  Can you provide a bit more context, or specific examples of what you'd like to see?

At the risk of over-explaining:  The HTTP protocol allows for both multiple headers and multiple URL query parameters with the same name.  This may be relatively rare, but it is entirely normal and compliant.  For go-httpbin — _unlike the original httpbin_ — we **do not** special-case the way we serialize single headers vs multiple headers or single query params vs multiple query params with a given name.  They are always serialized in go-httpbin's responses as arrays (single values as a 1-element array, multiple values as multiple-element arrays).

> Another way for this, use [] after key for list Ex:
> 
> ...
> 
> I know many frameworks use this way.

That is not part of the HTTP spec.  It is framework specific and entirely unnecessary, given a spec-compliant HTTP implementations.  I've never researched the history here, but I have always assumed that some frameworks made this implementation decision out of ignorance of the HTTP spec, but maybe there are historical reasons that necessitated this that I'm not aware of.

Regardless, go-httpbin is not going to implement any special cases like this.

(See https://github.com/mccutchen/go-httpbin/issues/37#issuecomment-770427446 for prior discussion of this `key[]=1&key[]=2` idea.)

PR #143, opened 2023-08-09
Upgrade Go to 1.21

harryzcy:


.github/workflows/test.yaml
```diff
@@ -43,6 +43,7 @@ jobs:
     strategy:
       matrix:
         go_version:
+        - '1.20'
         - '1.19'
         - '1.18'
```
mccutchen, line 48: Please drop the oldest release here, we just test against the current release (handled above on line 27) and the two previous releases.

mccutchen (commented):
This was quick, I saw this change before I saw that 1.21 was actually released!  Thanks for the contribution, just one small request below.

Note that I probably won't cut a new release immediately, so it may take a little bit for docker images built on 1.21 to be available.

PR #148, opened 2023-10-06
feat: special case CloudFlare client IP addrs

vanodevium:


httpbin/helpers.go
```diff
@@ -50,6 +50,11 @@ func getClientIP(r *http.Request) string {
 		return clientIP
 	}
 
+	// Special case Cloudflare for the value directly.
```
mccutchen, line 53: Style nit:  The comment above the previous conditional already explains this case, let's drop this comment and delete the blank line between the fly and cloudflare conditionals.

httpbin/helpers.go
```diff
@@ -50,6 +50,11 @@ func getClientIP(r *http.Request) string {
...
+	if cfClientIP := r.Header.Get("CF-Connecting-IP"); cfClientIP != "" {
+		return cfClientIP
```
mccutchen, line 55: Style nit:  please follow the pattern in the previous conditional and just call this `clientIP` rather than `cfClientIP`.

In case it's not obvious, in Golang's

```go
if x := y; test {}
```

declaration+conditional form, the declared variable does not escape the scope of the conditional, which means we don't need to worry about either shadowing the previous conditional's `clientIP` value or leaking this `clientIP` value into the rest of the function.

mccutchen (commented):
Thanks for the contribution, this makes sense to me.  Left a couple of very nit-picky style change requests below, let me know if they make sense to you!

mccutchen:
Oh, also, please add a test case for this to the existing unit tests for getClientIP in helpers_test.go.  (I expect our code coverage checks are gonna complain about this!)

vanodevium:
I have done this request from web editor :) Ok.
Please wait when I could work from laptop.

Your nits are awesome!

mccutchen (approved):
👍 LGTM pending CI run

vanodevium:
@mccutchen  Please tell when this functionality would be in docker image?

mccutchen:
Released in https://github.com/mccutchen/go-httpbin/releases/tag/v2.11.1.  A docker image for that release will be pushed as soon as [this workflow run finishes](https://github.com/mccutchen/go-httpbin/actions/runs/6433164965/job/17469649750).

mccutchen:
Docker image for `v2.11.1` now available, and `latest` tag points to the same:
https://hub.docker.com/r/mccutchen/go-httpbin/tags

PR #155, opened 2023-11-25
feat: add /websocket/echo endpoint

### LINKED ISSUE #151: feat: add `/ws` WebSockets endpoint
A `/ws` endpoint that provides a basic WebSockets-based echo server would be a useful addition, if it's not too difficult to implement without pulling in any external dependencies.  MDN's [guide to writing a WebSocket server](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_servers) might be a good starting point.

This would be very useful to help test HTTP proxies to ensure that they correctly support WebSockets, similar to the SSE endpoint proposed in #150.

We could potentially leverage something like https://github.com/crossbario/autobahn-testsuite to verify the basic functionality of the implementation, ideally wired up into our CI test suite.

mccutchen:
Here we add a new `/websocket/echo` endpoint, which implements a basic WebSocket echo service.

The endpoint is powered by our own basic, zero-dependency WebSocket implementation, which passes _almost_ every test in the invaluable [Autobahn Testsuite](https://github.com/crossbario/autobahn-testsuite) "fuzzingclient" set of integration tests, which will be run automatically as part of our continuous integration tests going forward.

The Autobahn Testsuite exceptions are documented here:
https://github.com/mccutchen/go-httpbin/blob/c7f36ed865b75fde01e5890e8f5d1137cc2377d4/httpbin/websocket/websocket_autobahn_test.go#L27-L38

Closes #151.


mccutchen:
Gonna land this, despite code coverage regressions.  We can follow up with expaneded websockets tests to improve some of the gaps.  In particular, would be good to test the frame and message size limit logic.

PR #160, opened 2023-12-12
feat: add /sse endpoint to test Server-Sent Events

### LINKED ISSUE #150: feat: add /sse endpoint to test Server-Sent Events
It's relatively easy to implement an endpoint that will use [Server-Sent Events](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events) to stream a response to a client, and it's also relatively easy to write HTTP proxy code that _does not_ correctly support SSE responses.

So, we should expose an endpoint in go-httpbin that makes it easier to test/validate that proxies correctly handle SSE.

mccutchen:
Each event is a "ping" that includes an incrementing integer ID and an integer Unix timestamp with millisecond resolution:

    event: ping
    data: {"id":9,"timestamp":1702417925258}

Fixes #150.

### DIFF
--- a/httpbin/httpbin.go
+++ b/httpbin/httpbin.go
@@ -1,6 +1,7 @@
 package httpbin
 
 import (
+	"bytes"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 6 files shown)


httpbin/httpbin.go
```diff
@@ -97,76 +124,59 @@ var _ http.Handler = &HTTPBin{}
...
-
+	mux.HandleFunc("/absolute-redirect/", h.AbsoluteRedirect)
 	mux.HandleFunc("/anything", h.Anything)
 	mux.HandleFunc("/anything/", h.Anything)
-
-	mux.HandleFunc("/ip", h.IP)
-	mux.HandleFunc("/user-agent", h.UserAgent)
-	mux.HandleFunc("/headers", h.Headers)
-	mux.HandleFunc("/response-headers", h.ResponseHeaders)
-	mux.HandleFunc("/hostname", h.Hostname)
```
mccutchen, line 119: Aside from adding the new `/sse` endpoint, this is just consolidating and alphabetizing the entries

PR #166, opened 2024-02-11
enable POST, PUT, DELETE and PATCH on basic auth

mgeuer:


mccutchen (commented):
Thanks!  Would you mind adding some test cases for this expanded functionality?

I think that could be as simple as wrapping just this "ok" test case in a loop over HTTP methods and running the same test code once for each method:
https://github.com/mccutchen/go-httpbin/blob/06f67278ff81cbcd09e9551d49c9f4dccdfe004d/httpbin/handlers_test.go#L1529-L1540

Roughly like so (ignoring whitespace changes):

```diff
diff --git a/httpbin/handlers_test.go b/httpbin/handlers_test.go
index 406ef03..b856f4d 100644
--- a/httpbin/handlers_test.go
+++ b/httpbin/handlers_test.go
@@ -1528,7 +1528,12 @@ func TestBasicAuth(t *testing.T) {
        t.Run("ok", func(t *testing.T) {
                t.Parallel()
 
-               req := newTestRequest(t, "GET", "/basic-auth/user/pass")
+               for _, method := range []string{"GET", "POST", "PUT", "DELETE", "PATCH"} {
+                       method := method
+                       t.Run(method, func(t *testing.T) {
+                               t.Parallel()
+
+                               req := newTestRequest(t, method, "/basic-auth/user/pass")
                                req.SetBasicAuth("user", "pass")
 
                                resp := must.DoReq(t, client, req)
@@ -1539,6 +1544,8 @@ func TestBasicAuth(t *testing.T) {
                                }
                                assert.DeepEqual(t, result, expectedResult, "expected authorized user")
                        })
+               }
+       })
```

mccutchen (approved):
👍 nice, thanks for the contribution!

PR #167, opened 2024-02-16
feat: support Fastly and Akamai headers for client IP addr

haccht:
Add support for Fastly and Akamai custom headres to get a reasonable value of the client ip address.


mccutchen (approved):
This is great, thanks for the contribution!

PR #169, opened 2024-02-27
add appProcotol to the k8s service for port name 'http'

bcollard:
Add the `appProcotol` field to the k8s service for port name 'http'

This is recommended - and sometimes required - by third-party apps like Istio or Kuma

Official doc: https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol


mccutchen:
Nice, thanks for the contribution!

Issue #74, opened 2022-03-31
[Feature request] Add prometheus or metrics api 


Lynskylate:

mccutchen:
Right now, this project does not have any dependencies outside the stdlib. I'm extremely reluctant to add any, in order to ensure that it continues to be as easy/safe as possible for other apps to include it in their testing dependencies without having to pull in a lot of additional 3rd party code.

That being said, it should be relatively easy to wrap this package and use the `Observer` hooks added in https://github.com/mccutchen/go-httpbin/pull/11 to add whatever extra instrumentation you need, though this does require building/deploying your own version.

Take a look at [examples/custom-instrumentation](https://github.com/mccutchen/go-httpbin/tree/main/examples/custom-instrumentation) to see what this might look like. (I just added this in https://github.com/mccutchen/go-httpbin/pull/75, because you reminded me that this was never actually documented!)

PR #11, opened 2019-01-15
Support pluggable logger

mccutchen:
This change allows users to specify their own logger for an `HTTPBin` instance, as long as it implements this minimal interface (which matches [the stdlib's `log.Logger`][1]):

```go
// Logger is a minimal logging interface
type Logger interface {
	Printf(format string, v ...interface{})
}
```

When using go-httpbin as a library (e.g. for unit tests), the default logger now _silently discards all messages_.

The `go-httpbin` binary & Docker container, on the other hand, will generate log messages like so:

```
./dist/go-httpbin
go-httpbin listening on 0.0.0.0:8080
time="2019-01-14T17:04:54.5177" status=200 method="GET" uri="/" size_bytes=11271 duration_ms=0.55
time="2019-01-14T17:04:59.048" status=418 method="GET" uri="/status/418" size_bytes=13 duration_ms=0.20
time="2019-01-14T17:05:03.1649" status=302 method="GET" uri="/redirect/6" size_bytes=0 duration_ms=0.04
time="2019-01-14T17:05:03.1754" status=302 method="GET" uri="/relative-redirect/5" size_bytes=0 duration_ms=0.02
time="2019-01-14T17:05:03.1826" status=302 method="GET" uri="/relative-redirect/4" size_bytes=0 duration_ms=0.03
time="2019-01-14T17:05:03.1908" status=302 method="GET" uri="/relative-redirect/3" size_bytes=0 duration_ms=0.02
time="2019-01-14T17:05:03.1961" status=302 method="GET" uri="/relative-redirect/2" size_bytes=0 duration_ms=0.02
time="2019-01-14T17:05:03.2042" status=302 method="GET" uri="/relative-redirect/1" size_bytes=0 duration_ms=0.03
time="2019-01-14T17:05:03.2113" status=200 method="GET" uri="/get" size_bytes=531 duration_ms=0.14
```

Fixes #9.

[1]: https://golang.org/pkg/log/#Logger.Printf

PR #179, opened 2024-08-20
feat: implement log formatting and fix some typos

### LINKED ISSUE #178: feat: implement log formatting
Would you consider adding new flag like `-log-format` to be able to configure logger handler for json logging or text logging?

pehlicd:
This pr aims to give option to select one of the log formats text or json. It also introduces new flag `-log-format` to achieve.

I modified the tests so it should work out of the box. Also fixed some typos that I saw in general.

Please let me know if you also want me to modify README.

Closes #178

### DIFF
--- a/httpbin/cmd/cmd.go
+++ b/httpbin/cmd/cmd.go
@@ -241,6 +241,9 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 			return nil, configErr("https cert and key must both be provided")
 		}
 	}
+	if cfg.LogFormat == "" && getEnv("LOG_FORMAT") != "" {
+		cfg.LogFormat = getEnv("LOG_FORMAT")
+	}
 	if cfg.LogFormat != "text" && cfg.LogFormat != "json" {
 		return nil, configErr("invalid log format %s, must be 'text' or 'json'", cfg.LogFormat)
 	}
--- a/httpbin/cmd/cmd.go
+++ b/httpbin/cmd/cmd.go
@@ -8,7 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"log"
+	"log/slog"
 	"net"
 	"net/http"
 	"os"
@@ -40,8 +40,6 @@ func Main() int {
 // mainImpl is the real implementation of Main(), extracted for better
 // testability.
 func mainImpl(args []string, getEnv func(string) string, getHostname func() (string, error), out io.Writer) int {
-	logger := log.New(out, "", 0)
-
 	cfg, err := loadConfig(args, getEnv, getHostname)
 	if err != nil {
 		if cfgErr, ok := err.(ConfigError); ok {
@@ -65,6 +63,14 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
 		}
 		fmt.Fprintf(out, "error: %s", err)
 		return 1
+	}
+
+	logger := slog.New(slog.NewTextHandler(out, nil))
+
+	if cfg.LogFormat == "json" {
+		// use structured logging if requested
+		handler := slog.NewJSONHandler(out, nil)
+		logger = slog.New(handler)
 	}
 
 	opts := []httpbin.OptionFunc{
@@ -93,7 +99,7 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
 	}
 
 	if err := listenAndServeGracefully(srv, cfg, logger); err != nil {
-		logger.Printf("error: %s", err)
+		logger.Error(fmt.Sprintf("error: %s", err))
 		return 1
 	}
 
@@ -113,6 +119,7 @@ type config struct {
 	RealHostname           string
 	TLSCertFile            string
 	TLSKeyFile             string
+	LogFormat              string
 
 	// temporary placeholders for arguments that need extra processing
 	rawAllowedRedirectDomains string
@@ -120,7 +127,7 @@ type config struct {
 }
 
 // ConfigError is used to signal an error with a command line argument or
-// environmment variable.
+// environment variable.
 //
 // It carries the command's usage output, so that we can decouple configuration
 // parsing from error reporting for better testability.
@@ -150,10 +157,11 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 	fs.StringVar(&cfg.TLSCertFile, "https-cert-file", "", "HTTPS Server certificate file")
 	fs.StringVar(&cfg.TLSKeyFile, "https-key-file", "", "HTTPS Server private key file")
 	fs.StringVar(&cfg.ExcludeHeaders, "exclude-headers", "", "Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard matching.")
+	fs.StringVar(&cfg.LogFormat, "log-format", "text", "Log format (text or json)")
 
 	// in order to fully control error output whether CLI arguments or env vars
 	// are used to configure the app, we need to take control away from the
-	// flagset, which by defaults prints errors automatically.
+	// flag-set, which by defaults prints errors automatically.
 	//
 	// so, we capture the "usage" output it would generate and then trick it
 	// into generating no output on errors, since they'll be handled by the
@@ -233,6 +241,9 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 			return nil, configErr("https cert and key must both be provided")
 		}
 	}
+	if cfg.LogFormat != "text" && cfg.LogFormat != "json" {
+		return nil, configErr("invalid log format %s, must be 'text' or 'json'", cfg.LogFormat)
+	}
 
 	// useRealHostname will be true if either the `-use-real-hostname`
 	// arg is given on the command line or if the USE_REAL_HOSTNAME env var
@@ -263,7 +274,7 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 	return cfg, nil
 }
 
-func listenAndServeGracefully(srv *http.Server, cfg *config, logger *log.Logger) error {
+func listenAndServeGracefully(srv *http.Server, cfg *config, logger *slog.Logger) error {
 	doneCh := make(chan error, 1)
 
 	go func() {
@@ -271,7 +282,7 @@ func listenAndServeGracefully(srv *http.Server, cfg *config, logger *log.Logger)
 		signal.Notify(sigCh, syscall.SIGTERM, syscall.SIGINT)
 		<-sigCh
 
-		logger.Printf("shutting down ...")
+		logger.Info("shutting down ...")
 		ctx, cancel := context.WithTimeout(context.Background(), cfg.MaxDuration+1*time.Second)
 		defer cancel()
 		doneCh <- srv.Shutdown(ctx)
@@ -279,10 +290,10 @@ func listenAndServeGracefully(srv *http.Server, cfg *config, logger *log.Logger)
 
 	var err error
 	if cfg.TLSCertFile != "" && cfg.TLSKeyFile != "" {
-		logger.Printf("go-httpbin listening on https://%s", srv.Addr)
+		logger.Info(fmt.Sprintf("go-httpbin listening on https://%s", srv.Addr))
 		err = srv.ListenAndServeTLS(cfg.TLSCertFile, cfg.TLSKeyFile)
 	} else {
-		logger.Printf("go-httpbin listening on http://%s", srv.Addr)
+		logger.Info(fmt.Sprintf("go-httpbin listening on http://%s", srv.Addr))
 		err = srv.ListenAndServe()
 	}
 	if err != nil && err != http.ErrServerClosed {
--- a/httpbin/cmd/cmd.go
+++ b/httpbin/cmd/cmd.go
@@ -24,6 +24,7 @@ import (
 const (
 	defaultListenHost = "0.0.0.0"
 	defaultListenPort = 8080
+	defaultLogFormat  = "text"
 
 	// Reasonable defaults for our http server
 	srvReadTimeout       = 5 * time.Second
@@ -157,7 +158,7 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 	fs.StringVar(&cfg.TLSCertFile, "https-cert-file", "", "HTTPS Server certificate file")
 	fs.StringVar(&cfg.TLSKeyFile, "https-key-file", "", "HTTPS Server private key file")
 	fs.StringVar(&cfg.ExcludeHeaders, "exclude-headers", "", "Drop platform-specific headers. Comma-separated list of headers key to drop, supporting wildcard matching.")
-	fs.StringVar(&cfg.LogFormat, "log-format", "text", "Log format (text or json)")
+	fs.StringVar(&cfg.LogFormat, "log-format", defaultLogFormat, "Log format (text or json)")
 
 	// in order to fully control error output whether CLI arguments or env vars
 	// are used to configure the app, we need to take control away from the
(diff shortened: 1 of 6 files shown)


mccutchen (commented):
Thanks for the contribution (and the typo fixes)!

In the past, I've been reluctant to make go-httpbin too opinionated about logging, given the wide variety of possible use cases.  Instead, we added a generic mechanism for pluggable _observers_ to enable users to add whatever instrumentation they needed (see https://github.com/mccutchen/go-httpbin/pull/11, https://github.com/mccutchen/go-httpbin/issues/74, etc).

However, now that `slog` is in the stdlib of the two most recent go versions it is now safe for us to adopt, and JSON structured logging seems "standardized enough" to warrant easy support here, so I think a change like this makes sense.

Left a couple of small comments below, plus I think there are some test failures to address, but LGTM at a high level!

httpbin/middleware.go
```diff
@@ -163,22 +163,19 @@ type Observer func(result Result)
...
-		)
+		l.With("time", time.Now().Format(dateFmt)).
+			With("status", result.Status).
+			With("method", result.Method).
+			With("uri", result.URI).
+			With("size_bytes", result.Size).
+			With("duration_ms", result.Duration.Seconds()*1e3).
+			With("user_agent", result.UserAgent).
+			With("client_ip", result.ClientIP).
+			Info("request")
```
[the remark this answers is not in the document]
pehlicd: Hey, thanks for the review and the insightful comments. Sorry for the late reply.

Regarding the suggestion to drop the request message:

I understand the point you're making, but it's a bit tricky with the `slog`. Unfortunately, with slog, there's no built-in way to log structured data without including a msg field when using the default handlers. I could set it as an empty string, but that looks awkward and doesn't seem like a good solution. Here's an example of what that would look like:

```
time=2024-08-26T00:54:24.005+02:00 level=INFO msg="" time=2024-08-26T00:54:24.0053 status=200 method=GET uri=/ size_bytes=11839 duration_ms=0.083083 user_agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Safari/537.36" client_ip=127.0.0.1:63536
```

If I understood correctly, you're suggesting parsing all the fields within the msg field for better readability, especially for tools like DataDog. However, in my experience with Elasticsearch, it's generally more effective and considered best practice to keep each field separate. This approach enhances searchability and performance.

Do you think there's room for a compromise, or should we stick with the current approach given the limitations?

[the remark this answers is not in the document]
mccutchen: > If I understood correctly, you're suggesting parsing all the fields within the msg field for better readability, especially for tools like DataDog. However, in my experience with Elasticsearch, it's generally more effective and considered best practice to keep each field separate. This approach enhances searchability and performance.

Not quite.  What I'm suggesting is that if we are going to include a message, we should make it useful on its own.  What we do at work is duplicate some of the structured fields into a human-friendly traditional web log style message, like so (using [zerolog](https://github.com/rs/zerolog) rather than slog):

```go
// Use DataDog's standard log attributes:
// https://docs.datadoghq.com/logs/processing/attributes_naming_convention/#default-standard-attribute-list
hlog.FromRequest(r).
	WithLevel(lvl).
	Int64("duration", duration.Nanoseconds()).
	Str("http.host", r.Host).
	Str("http.method", r.Method).
	Int("http.status_code", status).
	Str("http.url", r.URL.String()).
	Str("http.useragent", r.Header.Get("User-Agent"))).
	Int64("network.bytes_read", r.ContentLength).
	Int("network.bytes_written", size).
	Str("network.client.ip", getClientIP(r)))).
	Msgf("%d %s %s (%.1fms)", status, r.Method, r.RequestURI, durationMS)
```

So a log message might end up looking like

```
200 GET /example 12ms
```

where the status, method, request URI and duration are still included directly in the structured fields and can be queried/graphed/filtered/etc.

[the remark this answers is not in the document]
pehlicd: Thanks for clarification @mccutchen I was kinda confused about what you were asking. I hope now it should be like you wanted 🙂

httpbin/cmd/cmd.go
```diff
@@ -233,6 +241,9 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 			return nil, configErr("https cert and key must both be provided")
 		}
 	}
+	if cfg.LogFormat != "text" && cfg.LogFormat != "json" {
+		return nil, configErr("invalid log format %s, must be 'text' or 'json'", cfg.LogFormat)
+	}
```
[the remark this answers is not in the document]
pehlicd: Added env var check as well and updated the README

httpbin/cmd/cmd_test.go
```diff
@@ -472,6 +527,7 @@ func TestMainImpl(t *testing.T) {
 		getHostname func() (string, error)
 		wantCode    int
 		wantOut     string
+		regex       bool
```
pehlicd, line 530: I added this one for `server error` and `tls cert error` tests in the **TestMainImpl** because in these tests were failing because I was unable to mock slog without giving slog as input to `mainImpl` function. Instead of updating function it made sense for me to just check these two tests using regex.

httpbin/cmd/cmd.go
```diff
@@ -233,6 +241,12 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
 			return nil, configErr("https cert and key must both be provided")
 		}
 	}
+	if cfg.LogFormat == "" && getEnv("LOG_FORMAT") != "" {
```
mccutchen, line 244: This doesn't work as intended, because `cfg.LogFormat` will always have a default value of `"text"` based on the way the flag is defined above.

I think the best we can do is handle this like we do other config options with default values, where we check the env var when the value matches the default:
https://github.com/mccutchen/go-httpbin/blob/874932b8a03a74758ac8ef61439fcd4a0550211f/httpbin/cmd/cmd.go#L193-L201

It's not perfect, but it's good enough.

httpbin/cmd/cmd.go
```diff
@@ -233,6 +241,12 @@ func loadConfig(args []string, getEnv func(string) string, getHostname func() (s
...
+		cfg.LogFormat = getEnv("LOG_FORMAT")
+	}
+	if cfg.LogFormat != "text" && cfg.LogFormat != "json" {
+		return nil, configErr("invalid log format %s, must be 'text' or 'json'", cfg.LogFormat)
```
mccutchen, line 248: Nit:  Let's improve the formatting here:

```suggestion
		return nil, configErr(`invalid log format %q, must be "text" or "json"`, cfg.LogFormat)
```

httpbin/cmd/cmd_test.go
```diff
@@ -532,6 +595,13 @@ func TestMainImpl(t *testing.T) {
 				t.Fatalf("expected return code %d, got %d", tc.wantCode, gotCode)
 			}
 
+			if tc.regex {
+				if _, err := regexp.MatchString(tc.wantOut, out); err != nil {
```
mccutchen, line 599: This test doesn't work, we need to check the currently-ignored `matched` bool returned by regexp.MatchString.  When you do that, you'll see that the tests fail.

I think a better approach might be to have a separate `wantOutFn func(out string) bool` entry in the testcase struct, and for these couple of "regex" tests instead check for specific substrings?

mccutchen (commented):
A few more comments, but this is coming along!

httpbin/cmd/cmd_test.go
mccutchen: Yeah, this kind of test is going to be fairly fiddly to write with these logging changes.  Unfortunately, [as noted below](https://github.com/mccutchen/go-httpbin/pull/179#discussion_r1734660515), this approach doesn't actually work.

httpbin/middleware.go
```diff
@@ -163,22 +164,24 @@ type Observer func(result Result)
...
-			result.URI,
-			result.Size,
-			result.Duration.Seconds()*1e3, // https://github.com/golang/go/issues/5491#issuecomment-66079585
-			result.UserAgent,
-			result.ClientIP,
+		l.LogAttrs(
+			context.Background(),
+			slog.LevelInfo,
+			fmt.Sprintf(logFmt, result.Status, result.Method, result.URI, result.Duration.Seconds()*1e3),
+			slog.String("time", time.Now().Format(dateFmt)),
```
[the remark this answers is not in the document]
pehlicd: Yeah true, I will drop time field.

httpbin/cmd/cmd_test.go
```diff
@@ -431,6 +464,28 @@ func TestLoadConfig(t *testing.T) {
...
+			},
+		},
+		"ok use text log format": {
+			args: []string{"-log-format", "text"},
+			wantCfg: &config{
+				ListenHost:  "0.0.0.0",
+				ListenPort:  8080,
+				MaxBodySize: httpbin.DefaultMaxBodySize,
+				MaxDuration: httpbin.DefaultMaxDuration,
+				LogFormat:   testDefaultLogFormat,
```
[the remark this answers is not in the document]
mccutchen: No, just this specific one, where we're very explicitly testing a specific value.

httpbin/cmd/cmd.go
pehlicd: Ohh I see definitely, sorry for logic mistake. Will fix that as well!

httpbin/cmd/cmd_test.go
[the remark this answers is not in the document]
pehlicd: Got it thanks for clarification 🙏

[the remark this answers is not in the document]
mccutchen: … but thinking about it just a bit more, it would probably make sense to define this as a real default value in `cmd.go` the same way we do with these:
https://github.com/mccutchen/go-httpbin/blob/874932b8a03a74758ac8ef61439fcd4a0550211f/httpbin/cmd/cmd.go#L25-L26

Then, that real default var can be referenced here in the test suite rather than this separate `testDefaultLogFormat` var.

(This specific test should still test the specific value without using the variable, though.)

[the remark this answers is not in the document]
pehlicd: Yeah it definitely makes sense, I did the change on my local 👍

httpbin/cmd/cmd_test.go
pehlicd: Yeah I agree but I kinda lost about which texts I should actually check in sepreate `wantOutFn`. Could you please give me some example for that?

pehlicd: I pushed my changes about this one as well.

httpbin/cmd/cmd_test.go
```diff
@@ -431,6 +465,28 @@ func TestLoadConfig(t *testing.T) {
 				MaxBodySize:            httpbin.DefaultMaxBodySize,
 				MaxDuration:            httpbin.DefaultMaxDuration,
 				AllowedRedirectDomains: []string{"foo", "bar", "baz"},
+				LogFormat:              defaultLogFormat,
+			},
+		},
+		"ok use json log format": {
+			args: []string{"-log-format", "json"},
+			env:  map[string]string{"LOG_FORMAT": "json"},
```
mccutchen, line 473: Just noticed it, but this should be two separate tests, one for the CLI arg and one for the env var.  It would have caught the bug in processing the configuration I pointed out in an earlier review (where the LOG_FORMAT env var was ignored)!

httpbin/cmd/cmd_test.go
```diff
@@ -500,7 +557,25 @@ func TestMainImpl(t *testing.T) {
...
+
+				pattern := `(?m)^` + strings.Join(patternParts, `\n`)
+
+				matched, err := regexp.MatchString(pattern, out)
+				if err != nil {
+					fmt.Printf("Regex error: %v\n", err)
+					return false
+				}
+				return matched
+			},
```
mccutchen, line 578: Apologies for not being clear enough earlier, but I think this is more complex than is necessary.  I think each test case should specify either `wantOut` OR `wantOutFn`, but not both.  This test might be simplified down to something like:

```go
wantOutFn: func(t *testing.T, out string) bool {
    assert.Contains(t, out, `msg="error: listen tcp: address -256: invalid port"`, "error log"`)
},
```

I think that would accomplish the same thing but much more clearly.  (I'm using go-httpbin's own [internal assert package helper](https://github.com/mccutchen/go-httpbin/blob/main/internal/testing/assert/assert.go#L95-L101) here.)

--

This obviously implies a change to the `wantOutFn` signature in the testcase struct definition, and to the behavior of the actual test code calling the function, and to the similar testcase below.

I hope that makes sense overall!

httpbin/middleware.go
```diff
@@ -163,22 +164,27 @@ type Observer func(result Result)
...
-			result.URI,
-			result.Size,
-			result.Duration.Seconds()*1e3, // https://github.com/golang/go/issues/5491#issuecomment-66079585
-			result.UserAgent,
-			result.ClientIP,
+		if result.Status >= 500 {
+			logLevel = slog.LevelError
+		} else if result.Status >= 400 && result.Status < 500 {
+			logLevel = slog.LevelWarn
+		}
```
mccutchen, line 176: This isn't safe, we're modifying a `logLevel` var defined outside this function on every invocation.  I _think_ that means a single 5xx response followed by a 2xx response will be logged at error level.

I'd probably do something like this instead:

```go
logLevel := slog.InfoLevel
if result.Status >= 500 {
    logLevel = slog.LevelError
} else if result.Status >= 400 {
    logLevel = slog.LevelWarn
}
```

---

Yep, confirmed locally with a `/status/500` then `/status/200` request, where the latter is logged at ERROR level:

```
$ make run
mkdir -p dist
CGO_ENABLED=0 go build -ldflags="-s -w" -o dist/go-httpbin ./cmd/go-httpbin
HOST=127.0.0.1 PORT=8080 dist/go-httpbin
time=2024-08-28T14:18:06.611-04:00 level=INFO msg="go-httpbin listening on http://127.0.0.1:8080"
time=2024-08-28T14:18:14.150-04:00 level=ERROR msg="500 GET /status/500 0.0ms" status=500 method=GET uri=/status/500 size_bytes=0 duration_ms=0.01675 user_agent=curl/8.7.1 client_ip=127.0.0.1:60335
time=2024-08-28T14:18:24.516-04:00 level=ERROR msg="200 GET /status/200 0.1ms" status=200 method=GET uri=/status/200 size_bytes=0 duration_ms=0.08220799999999999 user_agent=curl/8.7.1 client_ip=127.0.0.1:60338
```

mccutchen (commented):
Apologies for how many rounds of review this is taking, I appreciate your patience.  A few, hopefully final, comments below.

httpbin/middleware.go
pehlicd: Oh yeah true, sorry about that, will fix it in my next commit.

httpbin/cmd/cmd_test.go
pehlicd: agreed will add separate test for that

pehlicd:
> Apologies for how many rounds of review this is taking, I appreciate your patience. A few, hopefully final, comments below.

No need to apologize 🙂. It's completely understandable that maintaining high-quality code requires thorough review.

httpbin/cmd/cmd_test.go
pehlicd: Yeah it makes sense, it was overcomplex which is not desirable, thanks a lot for showing me this way

httpbin/cmd/cmd_test.go
mccutchen: I'd be in favor of dropping this original test now, I don't think we get any real value of testing what happens when we configure the same logging via both env var and CLI flag

pehlicd: Yeah true, I dropped that. Also added one more case to test if nothing provided is it using default log format or not, I don't know if it makes sense for you

mccutchen:
Thanks for the contribution @pehlicd!  This is now released as [v2.14.1](https://github.com/mccutchen/go-httpbin/releases/tag/v2.14.1).

Issue #114, opened 2023-03-08
feat: add /env endpoint to allow exposing operator-controlled info from the server


liangyuanpeng:
It can return some env to user, like 

set HTTPBIN_VERSION=v1 for go-httpbin adn request /env, it can be return :

```json
{
  "envs":{
   "version":"v1"
  }
}
```

mccutchen:
Is the idea that this new `/env` endpoint would return all environment vars prefixed with `HTTPBIN_`?

I'm not sure I see a ton of value here, but I'm curious what use cases you have in mind.  Is this mainly a way to let the operator of a go-httpbin instance provide arbitrary data to clients?

mccutchen:
Gonna close this for now.  Feel free to re-open with more context!

charlesvhe:
Hi @mccutchen , I'm looking for this feature too.

My scenario is used to quickly verify gray release (traffic control), which requires differentiating between instances. For example, deploy go-httpbin instance A with ENV label=instanceA, instance B with ENV label=instanceB, and then implement traffic control through API gateway or Istio configuration. And verify the /env return to check the instances is right or not.

The current solution is to mount different images and access /image/jpeg to confirm which instance it is returning.

I think just return all environment vars  is OK.

mccutchen:
Thanks for the additional context @charlesvhe, I'll re-open this.  Are you interested in taking a pass at implementation?

> I think just return all environment vars is OK.

I would not do this by default, to minimize the risk of unintentionally leaking sensitive information.  Instead, we should only respond with env vars matching a particular prefix (e.g. `HTTPBIN_*` or `HTTPBIN_ENV_*`), so that operators may explicitly expose some env vars as necessary.

mloskot:
I've given this a go and drafted PR #189

PR #189, opened 2024-09-21
feat: add /env endpoint to allow exposing operator-controlled info from the server

mloskot:
Proposal of implementation of the feature request
- #114

----

I've submitted this for an initial round of review. 
I still have to add more cases to the test and add the endpoint to the index page, but first I'd like to receive feedback if this implementation goes in the right direction.

----

```bash
make run

curl http://127.0.0.1:8080/env
{
  "env": {}
}

export HTTPBIN_A=a
export HTTPBIN_B=b
export HTTPBIN_123=123

make run

curl http://127.0.0.1:8080/env
{
  "env": {
    "HTTPBIN_123": "123",
    "HTTPBIN_A": "a",
    "HTTPBIN_B": "b"
  }
}
```

### DIFF
--- a/httpbin/cmd/cmd.go
+++ b/httpbin/cmd/cmd.go
@@ -74,10 +74,20 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
 		logger = slog.New(handler)
 	}
 
+	env := make(map[string]string)
+	for _, e := range os.Environ() {
+		v := strings.SplitN(e, "=", 2)
+		if !strings.HasPrefix(v[0], "HTTPBIN_") {
+			continue
+		}
+		env[v[0]] = v[1]
+	}
+
 	opts := []httpbin.OptionFunc{
 		httpbin.WithMaxBodySize(cfg.MaxBodySize),
 		httpbin.WithMaxDuration(cfg.MaxDuration),
 		httpbin.WithObserver(httpbin.StdLogObserver(logger)),
+		httpbin.WithEnv(env),
 		httpbin.WithExcludeHeaders(cfg.ExcludeHeaders),
 	}
 	if cfg.Prefix != "" {
--- a/httpbin/cmd/cmd.go
+++ b/httpbin/cmd/cmd.go
@@ -25,6 +25,7 @@ const (
 	defaultListenHost = "0.0.0.0"
 	defaultListenPort = 8080
 	defaultLogFormat  = "text"
+	defaultEnvPrefix  = "HTTPBIN_ENV_"
 
 	// Reasonable defaults for our http server
 	srvReadTimeout       = 5 * time.Second
@@ -41,7 +42,7 @@ func Main() int {
 // mainImpl is the real implementation of Main(), extracted for better
 // testability.
 func mainImpl(args []string, getEnvVal func(string) string, getEnviron func() []string, getHostname func() (string, error), out io.Writer) int {
-	cfg, err := loadConfig(args, getEnvVal, getHostname)
+	cfg, err := loadConfig(args, getEnvVal, getEnviron, getHostname)
 	if err != nil {
 		if cfgErr, ok := err.(ConfigError); ok {
 			// for -h/-help, just print usage and exit without error
@@ -74,20 +75,11 @@ func mainImpl(args []string, getEnvVal func(string) string, getEnviron func() []
 		logger = slog.New(handler)
 	}
 
-	httpbinEnv := make(map[string]string)
-	for _, envVar := range getEnviron() {
-		name, value, _ := strings.Cut(envVar, "=")
-		if !strings.HasPrefix(name, "HTTPBIN_ENV_") {
-			continue
-		}
-		httpbinEnv[name] = value
-	}
-
 	opts := []httpbin.OptionFunc{
+		httpbin.WithEnv(cfg.Env),
 		httpbin.WithMaxBodySize(cfg.MaxBodySize),
 		httpbin.WithMaxDuration(cfg.MaxDuration),
 		httpbin.WithObserver(httpbin.StdLogObserver(logger)),
-		httpbin.WithEnv(httpbinEnv),
 		httpbin.WithExcludeHeaders(cfg.ExcludeHeaders),
 	}
 	if cfg.Prefix != "" {
@@ -120,6 +112,7 @@ func mainImpl(args []string, getEnvVal func(string) string, getEnviron func() []
 // config holds the configuration needed to initialize and run go-httpbin as a
 // standalone server.
 type config struct {
+	Env                    map[string]string
 	AllowedRedirectDomains []string
 	ListenHost             string
 	ExcludeHeaders         string
@@ -154,7 +147,7 @@ func (e ConfigError) Error() string {
 
 // loadConfig parses command line arguments and env vars into a fully resolved
 // Config struct. Command line arguments take precedence over env vars.
-func loadConfig(args []string, getEnv func(string) string, getHostname func() (string, error)) (*config, error) {
+func loadConfig(args []string, getEnvVal func(string) string, getEnviron func() []string, getHostname func() (string, error)) (*config, error) {
 	cfg := &config{}
 
 	fs := flag.NewFlagSet("go-httpbin", flag.ContinueOnError)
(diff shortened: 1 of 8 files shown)


httpbin/responses.go
```diff
@@ -13,6 +13,10 @@ const (
 	textContentType   = "text/plain; charset=utf-8"
 )
 
+type environmentResponse struct {
+	Environment map[string]string `json:"env"`
```
mccutchen, line 17: Naming nit:  The endpoint is `/env` and the key in the JSON response is `"env"`, let's call this `envResponse` and name the struct member `Env` to match?

mccutchen (commented):
Thanks for checking in on approach before investing too much work here!

While this implementation is nice and straightforward, I think I'd prefer a slightly different approach where the environment is resolved at process startup and passed explicitly to `httpbin.New()` via a new `httpbin.WithEnv(env map[string]string)` option func.

The general idea here is to do the work once instead of on every request when we can (sort of like how we pre-render all templates or only look up the hostname once rather than on every request).

Does that make sense to you?

mloskot:
@mccutchen 
> I think I'd prefer a slightly different approach where the environment is resolved at process startup and passed explicitly to `httpbin.New()` via a new `httpbin.WithEnv(env map[string]string)` option func.

This is a very good suggestion indeed. I will update this PR with the process startup approach.

httpbin/responses.go
mloskot: Sure, done

mloskot:
@mccutchen I've refatored and the `HTTPBIN_` vars are now collected at the startup. I'll welcome any feedback - I'm available to implement any suggestions you may have.

httpbin/cmd/cmd.go
```diff
@@ -74,10 +74,20 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
 		logger = slog.New(handler)
 	}
 
+	env := make(map[string]string)
+	for _, e := range os.Environ() {
+		v := strings.SplitN(e, "=", 2)
```
mccutchen, line 79: Minor, but I'd write this as

```go
key, val, _ := strings.Cut(e, "=")
```

just because I prefer giving things names rather than referring to opaque slice indexes where possible/convenient.

httpbin/cmd/cmd.go
```diff
@@ -74,10 +74,20 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
...
+		if !strings.HasPrefix(v[0], "HTTPBIN_") {
```
mccutchen, line 80: Here's something I'd like to think about a little bit.  A random collection of thoughts:
- I'd be inclined to look for a more specific `HTTPBIN_ENV_` prefix, to further reinforce that these special env vars will show up in the response from the `/env` endpoint.
  - Especially in cases where the serivce is deployed in Kubernetes, which by default exposes a ton of details about other services in the cluster with `${SERVICE_NAME}_`-prefixed env vars. (This might be overly cautious, but I try to be extra conservative when making changes that stand a chance of exposing new information unexpectedly!)
- I'd also consider stripping the prefix from the keys returned by `/env`, so `HTTPBIN_ENV_FOO=1` would end up being returned as `{"FOO": "1"}`
  - Would this behavior be convenient/useful or too magical/confusing?

As the implementer and a potential user of this feature, what do you think about these options?

httpbin/cmd/cmd.go
```diff
@@ -74,10 +74,20 @@ func mainImpl(args []string, getEnv func(string) string, getHostname func() (str
 		logger = slog.New(handler)
 	}
 
+	env := make(map[string]string)
+	for _, e := range os.Environ() {
```
mccutchen, line 78: To properly test this, I think we need to add a new argument to `mainImpl` like the existing `getEnv` arg to allow some indirection and avoid directly using `os.Environ()` here.

Maybe something like this (including a name change to disambiguate the two env-related helper functions)?

```diff
- func mainImpl(args []string, getEnv func(string) string, getHostname func() (string, error), out io.Writer) int {
+ func mainImpl(args []string, getEnvVal func(string) string, getEnviron func() []string, getHostname func() (string, error), out io.Writer) int {
```

Does that make sense?  It'll allow us to inject whatever environment we need to in unit tests to fully exercise this functionality.

mccutchen (commented):
A few more comments/questions/suggestions, please let me know what you think!

httpbin/cmd/cmd.go
mloskot: Good idea, such 'unpaking tuples' is also my preference, I've just failed to find Go-idiomatic way - I'm a Go newbie.
Done.

httpbin/cmd/cmd.go
mloskot: I have just replaced the prefix as you suggested. The Kubernetes case is a very good reason to use a more specific prefix indeed.

> Would this behavior be convenient/useful or too magical/confusing?

I 'feel' that such magic could lead to confusion. 
What if an operator exports container with:

```
BAR=bar
FOO=foo
HTTPBIN_ENV_FOO=foo-or-goo
```

then it could lead to confusion like "Which `FOO` is listed by /env?"

I usually prefer when systems behave less magical and in more explicit way. So, I wouldn't mind seeing seemingly redundant `HTTPBIN_ENV_` prefix in the `/env` response.

mloskot: > (...) it could lead to confusion like "Which FOO is listed by /env?"

Or, we could make the env vars prefix configurable with `HTTPBIN_ENV_` as default stored in the constants

https://github.com/mccutchen/go-httpbin/blob/34a21a3ef81df5aa42185fcae2e5a608404ac6a0/httpbin/httpbin.go#L9-L14

Then, provide `--env-var-prefix` command line argument to let operators override it with desired prefix.

httpbin/cmd/cmd.go
mloskot: I've taken first stab at this idea, which is a good idea:
- modified `mainImpl` as suggested
- updated `TestMainImpl` accordingly

There are no test cases of `/env` endpoint handler that make use of this new `mainImpl` feature yet.

httpbin/cmd/cmd_test.go
```diff
@@ -53,8 +53,9 @@ func TestLoadConfig(t *testing.T) {
 
 	testCases := map[string]struct {
 		args        []string
-		env         map[string]string
+		optionsEnv  map[string]string
 		getHostname func() (string, error)
+		env         []string
```
mccutchen, line 58: This change adds a lot of unnecessary noise to the diff, I'd rather we take a different approach here.  Since we're already defining an env for these test cases, let's just use that env for both our `getEnvVal` and `getEnviron` stubs?

I.e. do something like this below:

```diff
- gotCode := mainImpl(tc.args, func(key string) string { return tc.optionsEnv[key] }, func() []string { return tc.env }, tc.getHostname, buf)
+ gotCode := mainImpl(tc.args, func(key string) string { return tc.env[key] }, func() []string { return environSlice(tc.env) }, tc.getHostname, buf)
```

where `environSlice` is a simple little helper to convert map[string]string into a `[]string` slice of key=val pairs.

httpbin/static/index.html.tmpl
```diff
@@ -82,6 +82,7 @@
 <li><a href="{{.Prefix}}/drip?code=200&amp;numbytes=5&amp;duration=5"><code>{{.Prefix}}/drip?numbytes=n&amp;duration=s&amp;delay=s&amp;code=code</code></a> Drips data over the given duration after an optional initial delay, simulating a slow HTTP server.</li>
 <li><a href="{{.Prefix}}/dump/request"><code>{{.Prefix}}/dump/request</code></a> Returns the given request in its HTTP/1.x wire approximate representation.</li>
 <li><a href="{{.Prefix}}/encoding/utf8"><code>{{.Prefix}}/encoding/utf8</code></a> Returns page containing UTF-8 data.</li>
+<li><a href="{{.Prefix}}/env"><code>{{.Prefix}}/env</code></a> Returns all environment variables named with <code>HTTPBIN_</code> prefix.</li>
```
mccutchen, line 85:
```suggestion
<li><a href="{{.Prefix}}/env"><code>{{.Prefix}}/env</code></a> Returns all environment variables named with <code>HTTPBIN_ENV_</code> prefix.</li>
```

mccutchen (commented):
Thanks for your persistence with this, it looks good to me overall, just a couple of hopefully final comments below.

It does look like we need a test for properly parsing the environment in `mainImpl`, I don't think I see any at a glance?

httpbin/cmd/cmd_test.go
mloskot: > It does look like we need a test for properly parsing the environment in mainImpl, I don't think I see any at a glance?

Yes, correct, I haven't managed to figure how to utilise the `mainImpl` for testing of the `/env` yet with some prefabricated `HTTPBIN_ENV_*` variables.
Perhaps, the `HTTPBIN_ENV_*` environ should become a member of the `config`, then the testing would become part of `loadConfig` test cases.
I'll welcome any suggestions. 

> This change adds a lot of unnecessary noise to the diff (...) where environSlice is a simple little helper (...)

I have reverted that `env` and `optionsEnv` splitting and implemented your suggestion.

mccutchen: > Yes, correct, I haven't managed to figure how to utilise the `mainImpl` for testing of the `/env` yet with some prefabricated `HTTPBIN_ENV_*` variables.
> Perhaps, the `HTTPBIN_ENV_*` environ should become a member of the `config`, then the testing would become part of `loadConfig` test cases.

Ahhhhh yeah, I think it might be worth it to move this new env configuration onto the config struct, just to make it testable.  Would you mind making that change?  After that, I think we're good to go here!

httpbin/cmd/cmd.go
mccutchen: Thanks for the input, I appreciate it!  For now, let's just keep this simple and use the `HTTPBIN_ENV_*` vars as we find them, without stripping the prefix.

httpbin/cmd/cmd_test.go
mloskot: > Would you mind making that change?

Yes, of course, I will continue near the end of this week.

mloskot: @mccutchen As promised, I've just reworked this PR to store the env in the config and expose it via HTTPBin, and covered it with tests based on `loadConfig` test cases.

I don't mind to keep working on polishing this PR until you're happy to merge it, no rush with accepting it.

mccutchen (approved):
👍 this came out really nice, thanks so much for your patience and persistence!

mloskot:
Pleasure is mine, thanks!

Issue #190, opened 2024-09-21
VERSION in Makefile resolved to Git hash instead of tag


mloskot:
The `git rev-parse --short HEAD` below resolves to Git hash 

https://github.com/mccutchen/go-httpbin/blob/34a21a3ef81df5aa42185fcae2e5a608404ac6a0/Makefile#L3-L4

Shouldn't the `VERSION` be based on Git tag, that is, `git describe --tags --abbrev=0`?

This makes me wonder, is this `VERSION` actually used for the image tagging?

Perhaps Docker images are tagged following this instruction?

https://github.com/mccutchen/go-httpbin/blob/34a21a3ef81df5aa42185fcae2e5a608404ac6a0/DEVELOPMENT.md?plain=1#L93-L95

If my understanding is correct and there is room for improvement indeed, I can propose PR.

mccutchen:
Fixed in #248

mloskot:
Thank you!

Issue #188, opened 2024-09-20
feat: Add /version endpoint and command line argument -version


mloskot:
Is there any reason that `go-httpbin` binary does not seem to offer any way to report its version?

I there isn't any, I'd like to suggest adding it, if possible.

Additionally, it would be nice to have `/version` endpoint.

I often have to test number of environments, clusters, DNS zones/domains, where I deploy different versions of go-httpbin, and being able to request the release version would be useful as a way to identify certain deployments.

If there is an interest, I'm keen in giving it a shot to implement it.

mccutchen:
No particular reason. Both additions sound like a good idea to me, I’ll gladly review a pull request if you open one!

mloskot:
Good to hear, I've proposed #191

mccutchen:
Fixed in #248

PR #191, opened 2024-09-21
feat: Add /version endpoint reporting build and version info

mloskot:
Refs:
- #188
- #190

Signed-off-by: Mateusz Łoskot <mateusz@loskot.net>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,13 @@
+BUILD_DATE       := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+BUILD_GIT_COMMIT := $(shell git rev-parse --short HEAD)
+BUILD_GIT_TAG    := $(shell git describe --tags --abbrev=0 | tr -d '\n')
+BUILD_LDFLAGS    := -X 'github.com/mccutchen/go-httpbin/v2/httpbin.buildDate=$(BUILD_DATE)' -X 'github.com/mccutchen/go-httpbin/v2/httpbin.buildGitCommit=$(BUILD_GIT_COMMIT)' -X 'github.com/mccutchen/go-httpbin/v2/httpbin.buildGitTag=$(BUILD_GIT_TAG)'
+
 # The version that will be used in docker tags (e.g. to push a
 # go-httpbin:latest image use `make imagepush VERSION=latest)`
-VERSION    ?= $(shell git rev-parse --short HEAD)
+VERSION    ?= $(BUILD_GIT_TAG)
 DOCKER_TAG ?= mccutchen/go-httpbin:$(VERSION)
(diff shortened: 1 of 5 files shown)


mloskot:
Thank you @mccutchen for the extensive explanation - I have learned a lot.

When I started working on the idea of `/version` my first thought was "What is the canonical golang way?" and [I learned about the ldflags and `-X`](https://www.neosync.dev/blog/setting-go-version-information), got excited and submitted this PR.

However, perhaps a more explicit and old-school way would be appropriate to handle all the cases you mention, i.e, `VERSION` file which is read by `Makefile`. 

For example,

- tag `v2.15.0` has `VERSION` file containing `2.15.0`
- main branch has `VERSION` file containing `dev` or `main`

mccutchen:
> However, perhaps a more explicit and old-school way would be appropriate to handle all the cases you mention, i.e. `VERSION` file which is read by `Makefile`.
> 
> For example,
> 
>     * tag `v2.15.0` has `VERSION` file containing `2.15.0`
> 
>     * main branch has `VERSION` file containing `dev` or `main`

I'm not sure I like that approach either, just because it will be tricky to juggle different contents of a VERSION file and make it more complicated to cut a new release.  (What would that look like, one commit to bump VERSION, tag release, follow-up commit to revert VERSION back to `dev` or `main`?)

I'm sorry, while I do think this would be a good feature to add, I still don't quite have a clear idea in my own head for how we might implement it in a way that doesn't add too much complexity/burden to the development and release process.  I'll think about it a bit more; happy to keep working through ideas here if you come up with other approaches!

mloskot:
@mccutchen I perfectly understand your reluctance as the tradeoffs or complications are clear and obvious.

I'm just thinking out loud that Go ecosystem must have solved such problem of setting version in a way that is nicely integrated with Git workflow, binaries build and container images build workflows.

By the way, I've been looking for inspirations what other established Go projects do, e.g. Flux CD seems to rely on goreleaser which relies on tagging, so AFAIU everything starts from creating a semver-based tag:

- https://github.com/fluxcd/flux2/blob/cf26cf25f149e864b17598c0d0125f424e7bd25a/cmd/flux/main.go#L41
- https://github.com/fluxcd/flux2/blob/cf26cf25f149e864b17598c0d0125f424e7bd25a/.goreleaser.yml#L7
- https://github.com/fluxcd/flux2/blob/cf26cf25f149e864b17598c0d0125f424e7bd25a/Makefile#L1

while building from non-relased snapshot sets the so called dev version

- https://github.com/fluxcd/flux2/blob/cf26cf25f149e864b17598c0d0125f424e7bd25a/Makefile#L2

PR #209, opened 2025-05-01
feat: Add GitHub Actions workflow and Makefile support for multi-plat…

takumin:
# About

- Introduced `.github/workflows/build.yaml` for CI/CD, including build, checksum, and publish jobs
  - Supports Go cross-compilation for multiple OS/arch targets
  - Includes artifact attestation and checksum generation
  - Publishes artifacts to GitHub Releases on tag push
- Extended `Makefile` with `release` target to generate static binaries per target
  - Supports GOOS/GOARCH/GOARM detection and naming

# Notes

To check the operation, we issued a test release in the following repository and confirmed that it works properly.

Repository:
https://github.com/takumin/go-httpbin

Commit:
https://github.com/takumin/go-httpbin/commit/857d10fd453d2d9a71854c20a9968cc1bedec67f

GitHub Action Workflow Log:
https://github.com/takumin/go-httpbin/actions/runs/14769067526

GitHub Release:
https://github.com/takumin/go-httpbin/releases/tag/v2.18.2

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,171 @@
+---
+name: build
+
+on:
+  push:
+    branches: [main] # pushes TO main
+  pull_request:
+    branches: [main] # pull requests AGAINST main
+  release:
+    types: [published] # for release build
+
+# cancel CI runs when a new commit is pushed to any branch except main
+concurrency:
+  group: 'build-${{ github.ref }}'
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+
+    permissions:
+      contents: read # for uses: actions/checkout
+      id-token: write # for actions/attest-build-provenance
+      attestations: write # for actions/attest-build-provenance
+
+    strategy:
+      fail-fast: false
+      matrix:
+        # Static binaries are built with the `stable` version
+        # If you want to add a target, add it below.
+        include:
+        - {goos: 'darwin', goarch: 'amd64'}
+        - {goos: 'darwin', goarch: 'arm64'}
+        - {goos: 'linux', goarch: 'amd64'}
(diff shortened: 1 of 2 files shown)


mccutchen:
Thanks for the contribution, this kind of prebuilt binary release process is long overdue!

Before I dive too far into the code itself, I have a couple of high level questions:
- Why is the new workflow split into separate `build`/`checksum`/`publish` jobs?  It's not obvious to me at a glance why this isn't one job, but if I had to guess I'd wonder whether it's simpler to separate the build job (with its large matrix) from the rest of the steps, which can operate on all of the built binaries at once without needing to duplicate the matrix.
- Did you consider using [GoReleaser](https://goreleaser.com/) for this?  That's what I've had my eye on for a while, with their [example-supply-chain](https://github.com/goreleaser/example-supply-chain) repo as an enticing example.  Basically, I think we could do everything you've got here and more with a declarative configuration and off-the-shelf GH Action workflows.

mccutchen:
One more high level question:
- Does this build binaries for every commit on a pull request and/or every push to the `main` branch?   It seems like it _might_, but it's not obvious to me.  I'd rather just build binaries like this when a new release is tagged, and leave it to intrepid developers to build themselves if they need a pre-release version.

mccutchen:
Thanks for the detailed reply, I appreciate your thoughts here. 

I hope to dig in this weekend, I’ll keep you posted!

mccutchen (commented):
Hi @takumin, I'm sorry for the delay in taking a closer look at this.  It looks good to me overall, but I have left handful of comments and questions below.  Please let me know what you think!

mccutchen:
👋 hey @takumin, I saw the mention over on https://github.com/itamae-kitchen/mitamae/discussions/143#discussioncomment-17773352 and it made me realize I never really closed the loop here.

I just want to say a big **thank you** for kickstarting this process and demonstrating the various moving parts to figure out, I'm not sure I would have gotten around to #244 without this motivation!  So even though I ended up going a different direction, you laid the groundwork here, and I really appreciate it!

PR #213, opened 2025-06-11
fix: actually test fix for local port number in `/ip` response

mccutchen:
Ugh, forgot I had auto-merge turned on for #212, so it went in before I could push a test.


mccutchen:
(Playing with fire, auto-merging again)

PR #224, opened 2025-10-29
docker: switch base image from distroless/base to distroless/static:nonroot

yosida95:
go-httpbin is built fully statically and does not depend on libc, so it can use the distroless/static image. This change reduces the resulting image size by about half.

```console
$ docker image ls
REPOSITORY                     TAG       IMAGE ID       CREATED          SIZE
go-httpbin                     nonroot   9910daab0040   38 minutes ago   19.5MB
ghcr.io/mccutchen/go-httpbin   2.18.3    115573352e3d   4 months ago     38.1MB
```

Since go-httpbin's default listening port is 8080, an unprivileged port, it can run as non-root. While this may break setups that use privileged ports, running as non-root provides a clear security advantage. Such users can continue using their ports by adding the CAP_NET_BIND_SERVICE capability.

### DIFF
--- a/httpbin/websocket/websocket_autobahn_test.go
+++ b/httpbin/websocket/websocket_autobahn_test.go
@@ -19,7 +19,7 @@ import (
 	"github.com/mccutchen/go-httpbin/v2/internal/testing/assert"
 )
 
-const autobahnImage = "crossbario/autobahn-testsuite:0.8.2"
+const autobahnImage = "crossbario/autobahn-testsuite:25.10.1"
 
 var defaultIncludedTestCases = []string{
 	"*",


mccutchen:
This looks great, thanks!

I love the thoroughness of the README update, but I might actually push a small commit to link out to documentation on CAP_NET_BIND_SERVICE to keep it brief, given that this is unlikely — I assume — to be a common need.

(Also, I'll fix the test failure separately.)

mccutchen:
> I love the thoroughness of the README update, but I might actually push a small commit to link out to documentation on CAP_NET_BIND_SERVICE to keep it brief, given that this is unlikely — I assume — to be a common need.

On second thought, I kept the more detailed info but moved it down below.  I hope you don't mind!

Thanks a lot for the contribution, this is a nice little improvement to both image size and out-of-the-box security posture.

Issue #113, opened 2023-03-08
API /IP support server ip


liangyuanpeng:
API `/IP` return server ip, so it shoud be :  

```
{
"origin":"xxx",
"server":"172.0.2.1"
}
```

PR #226, opened 2025-11-18
feat: add `/server-ip` endpoint to return server's local IP address

paimus:
This change introduces a new /server-ip endpoint that returns the IP address of the server (or Pod in Kubernetes) handling the request, as opposed to the existing /ip endpoint which returns the client's IP address.

Changes:
- Add getServerIP() helper in helpers.go to detect server's primary IP
  - Prefers non-loopback IPv4 addresses
  - Falls back to IPv6 if no IPv4 found
  - Uses UDP dial as last resort for IP detection
- Add ServerIP handler in handlers.go
- Register /server-ip route in httpbin.go
- Add serverIPResponse struct in responses.go

This is useful for debugging and service discovery in containerized environments where the server's IP may need to be identified.

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -166,6 +166,13 @@ func (h *HTTPBin) Headers(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+// ServerIP echoes the IP address of the server handling the request
+func (h *HTTPBin) ServerIP(w http.ResponseWriter, _ *http.Request) {
+	writeJSON(http.StatusOK, w, &serverIPResponse{
+		ServerIP: getServerIP(),
+	})
+}
+
 type statusCase struct {
 	headers map[string]string
 	body    []byte
--- a/httpbin/helpers.go
+++ b/httpbin/helpers.go
@@ -106,6 +106,38 @@ func getURL(r *http.Request) *url.URL {
 	}
 }
 
+// getServerIP attempts to determine the current server's primary IP address
+// (e.g., Pod IP in Kubernetes). It prefers non-loopback IPv4 addresses, then
+// falls back to non-loopback IPv6 if necessary.
+func getServerIP() string {
+	// Prefer non-loopback IPv4 from interface addresses
+	if addrs, err := net.InterfaceAddrs(); err == nil {
+		var v6Fallback string
+		for _, a := range addrs {
+			if ipnet, ok := a.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+				ip := ipnet.IP
+				if v4 := ip.To4(); v4 != nil {
+					return v4.String()
+				}
+				if v6Fallback == "" {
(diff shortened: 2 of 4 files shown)


mccutchen:
I appreciate the contribution, but as noted in https://github.com/mccutchen/go-httpbin/issues/113#issuecomment-1466225358 I am reluctant to expose this _potentially_ sensitive info about internal networks by default.

I'd consider a change that only exposed this endpoint if it was explicitly enabled, along the lines of the `USE_REAL_HOSTNAME` option added in https://github.com/mccutchen/go-httpbin/pull/81.

That being said, I'm not sure about the implementation here.  Why is IPv4 preferred over IPv6?  Also I don't think making a throwaway UDP connection (why to port 80? does the port just not matter here?) is a reasonable thing to do.

mccutchen:
> First off UDP is a connectionless protocol.
> 
> `net.Dial("udp", "8.8.8.8:80")` -> This asks the kernel, “If I wanted to send a UDP packet to 8.8.8.8, which IP/interface would you use?”
> 
> Then you use `localAddr := conn.LocalAddr().(*net.UDPAddr)` to receive the local IP of the interface used. So no packets are actually sent unless you "Write" to the socket.

Ah, thank you for the explanation, TIL!

> I would may consider just making a json payload that shows everything

I think if we add this functionality I like this direction:  return all the IP addresses we know about.  Is there a reasonable heuristic we can use to choose a "primary" address that would make sense to most clients of this hypothetical endpoint?  If not, I'd be inclined to just return them all and let the client choose one.

mccutchen (changes requested):
Okay, to reset here:  I'd be willing to include this new endpoint with a few adjustments to the implementation:
- [ ] Make it opt-in, along the lines of the `USE_REAL_HOSTNAME` option added in https://github.com/mccutchen/go-httpbin/pull/81.
- [ ] Make it return a more thorough response like the one suggested by @derekmpage [above](https://github.com/mccutchen/go-httpbin/pull/226#issuecomment-3720464694) (lightly adapted here):
    ```json
    {
      "primary_address": {
        "10.0.5.25"
      },
      "addresses": {
        "ipv4": [
          "192.168.1.101",
          "10.0.5.25",
          "172.16.254.1"
        ],
        "ipv6": [
          "2001:0db8:85a3::8a2e:0370:7334",
          "fe80::1ff:fe23:4567:890a"
        ]
      }
    }
    ```
- [ ] where the primary address is chosen as described by @aarnaud [above](https://github.com/mccutchen/go-httpbin/pull/226#issuecomment-3805153726): "It's the local IP address that the kernel selects as the source address when routing traffic through the default gateway"

mccutchen:
Appreciate the discussion here but I'm going to close this due to inactivity. Happy to revisit if the suggestions in https://github.com/mccutchen/go-httpbin/pull/226#pullrequestreview-3726249695 make sense to folks.

Issue #242, opened 2026-03-15
release: Could we build and release the static executable binaries in GitHub Release for major platforms


lqhuang:
Hi Will,

Thanks for creating a modern version of `httpbin`! I like it and use it often.

I've encountered a situation where I'd rather avoid using Docker or Docker-in-Docker (for example, in certain CI actions). I also don't want to set up the Go toolchain to build the CLI from scratch. So, I was hoping to find pre-built binaries that I could download directly from the GitHub Releases. I checked the page, but it seems only source code is currently available.

Here are a few Go-based projects for reference:

- https://github.com/antonmedv/fx/releases/
- https://github.com/mikefarah/yq/releases/

which should help clarify my expected results and assets.

Regarding the primary platforms, I believe Windows x86_64, Linux x86_64/aarch64, and macOS aarch64 should be enough to cover most use cases in the first few experimental releases.

I don't have much experience with Go development, but if you're short on time, I'd be happy to help improve the GitHub Actions CI!

Looking forward to hearing your thoughts.

PS: Could we also add an echo server with `Content-Encoding` support via `/echo` like `/websocket/echo`?

Regards,
Lanqing

mccutchen:
Yes, I plan to do this!

When I can, I'm currently experimenting with a goreleaser based process over in https://github.com/mccutchen/go-app-template that I'm hoping to import here once I've ironed out the kinks.

mccutchen:
Fixed in #244.  As of [v2.22.0](https://github.com/mccutchen/go-httpbin/releases/tag/v2.22.0), each release will include prebuilt binaries for Linux, macOS, and Windows, for both amd64 and arm64 CPUs.  Thanks for the push!

lqhuang:
Thanks! I really appreciate it!

Cheers!

Issue #85, opened 2022-10-13
Consider switching from CMD to ENTRYPOINT in docker image


mccutchen:
As proposed by @x70b1 in https://github.com/mccutchen/go-httpbin/pull/84, updating the Dockerfile to set the `go-httpbin` binary as the `ENTRYPOINT` rather than the `CMD`

```diff
- CMD ["/bin/go-httpbin"]
+ ENTRYPOINT ["/bin/go-httpbin"]
```

would be a small, useful UX improvement by allowing folks who run the image using command line arguments to slightly simplify their workflows:

```diff
-  nerdctl run mccutchen/go-httpbin go-httpbin -use-real-hostname
+  nerdctl run mccutchen/go-httpbin -use-real-hostname
```

I'm currently a little reluctant to do this out of fear of unexpectedly breaking existing uses in CI processes if they're not pinning to a specific release, but it's a good candidate to include in any future breaking change we need to make.

mccutchen:
Done in #244 (and it turns out to be backwards compatible!)

PR #244, opened 2026-04-08
ci: updated release process

mccutchen:
Switch over to a [goreleaser][] based process, still triggered by the creation of a release on GitHub, which now
- Builds signed OCI images w/ SBOMs and pushes them to GHCR and Dockker Hub (same as before)
- Builds signed release binaries for Linux, Mac (w/ notarization), and Windows on both amd64 and arm64 architectures
- Generates build provenance attestations on GitHub

This closes #209, #242 and #85 (kinda incidentally).  It also lays the ground work for addressing #188 and #190, which will be fixed in follow ups.

My plan is to merge this and create pre-releases while I nail down any loose ends before cutting a final release.

[goreleaser]: https://github.com/goreleaser/goreleaser

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,75 +1,94 @@
-# This action pushes new amd64 and arm64 docker images to the Docker and GitHub
-# registries on every new release of the project.
+# The release workflow is triggered by the creation of a GitHub release and
+# builds cross-platform binaries and multi-arch docker images.
 #
-# Cobbled together from these sources:
-# - https://github.com/docker/build-push-action/#usage
-# - https://docs.github.com/en/actions/publishing-packages/publishing-docker-images
-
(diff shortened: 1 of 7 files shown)


nightah:
Is there a reason the default port `EXPOSE` was omitted in the release Dockerfile?
This has caused issues with Traefik, where it can no longer auto-discover the port for the Docker provider.

mccutchen:
Oh no! That was an oversight, let me push a fix. Thanks for the heads up!

mccutchen:
Should now be fixed in [v2.22.1](https://github.com/mccutchen/go-httpbin/releases/tag/v2.22.1) and `latest` images on ghcr and docker hub.  Thanks again!

nightah:
Thanks for the prompt response and fix.

PR #249, opened 2026-04-12
compat: avoid OCI entrypoint change breaking legacy invocations

mccutchen:
Before #244, the entrypoint for Docker/OCI images was a shell, so it was possible to run go-httpbin like so to pass CLI args:

    docker run mccutchen/go-httpbin go-httpbin -use-real-hostname

After that change, the entrypoint is the go-httpbin binary itself, so the above would become:

    docker run mccutchen/go-httpbin -use-real-hostname

On the off chance there are users who a) rely on the first form and b) also automatically update to the latest image, we can pretty easily add a bit of shim code to allow the old invocation to work in new images.


mccutchen:
Heh, actually, turns out this isn't really needed, as positional args seem to be ignored entirely!

```
./dist/go-httpbin xgo-httpbin asdfasd adfasfasdf -use-real-hostname
time=2026-04-11T22:41:35.544-04:00 level=INFO msg="go-httpbin listening on http://0.0.0.0:8080"
```

PR #255, opened 2026-05-20
feat: support text format for /ip

### LINKED ISSUE #254: feat: support plain text responses from `/ip`
Hi there

Quite happy to discover this httpbin alternative, thanks a lot. 

It covered my original use case, and I saw an interesting /ip (actually exists in httpbin, never paid attention). Would happily replace my https://icanhazip.com/ usage by it, but would it be possible to add a parameter like `?format=text`, or maybe a dedicated /ip_raw endpoint to have the raw IP instead of JSON ? Otherwise in my scripts I'd have to either add jq (quite overweight) or something like `| grep | cut -d'"' -f2`, kinda brittle.

Thanks again,

nanookclaw:
## Summary

Adds support for `GET /ip?format=text`, returning the detected client IP directly as `text/plain; charset=utf-8`. The default `/ip` response stays unchanged and continues to return `{"origin": ...}` JSON.

Closes #254.

## Testing

- `go test ./httpbin -run 'TestIP/text_format$' -count=1`
- `go test ./httpbin -run 'TestIP/(remote_addr_used_if_no_x-forwarded-for|first_entry_in_x-forwarded-for_used_if_present)$' -count=1`
- `go test ./httpbin -run '^$' -count=1`
- `git diff --check`

Note: `go test ./httpbin -run TestIP -count=1` reaches the existing real-connection subtest and fails in this sandbox because `httptest` cannot listen on localhost (`operation not permitted`).

### DIFF
--- a/httpbin/handlers.go
+++ b/httpbin/handlers.go
@@ -170,8 +170,14 @@ func (h *HTTPBin) Deflate(w http.ResponseWriter, r *http.Request) {
 
 // IP echoes the IP address of the incoming request
 func (h *HTTPBin) IP(w http.ResponseWriter, r *http.Request) {
+	origin := getClientIP(r)
+	if r.URL.Query().Get("format") == "text" {
+		writeResponse(w, http.StatusOK, textContentType, []byte(origin))
+		return
+	}
+
 	writeJSON(http.StatusOK, w, &ipResponse{
-		Origin: getClientIP(r),
+		Origin: origin,
 	})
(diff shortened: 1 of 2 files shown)


mccutchen:
While there’s no official policy on AI-written contributions, I’m gonna reject this because I assume this is a bot that’s querying for “help wanted” labels and autosubmitting fixes. Not interested in this kind of contribution.

PR #261, opened 2026-06-07
build: fix static linking in distroless dockerfile

ajithcnambiar:
## Summary
  - Pin Docker build image to `golang:1.26.4` (was `golang:1.26`) for the latest stable Go.
  - Fix `Makefile` build target: drop the stray `;` after `CGO_ENABLED=0`. The semicolon made it a non-exported shell var, so `go
  build` defaulted to CGO enabled and produced a dynamically-linked binary — which fails on distroless with `exec /bin/go-httpbin:
  no such file or directory`. Now statically linked.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.26.4 AS build
+FROM golang:1.26 AS build
 
 WORKDIR /go/src/github.com/mccutchen/go-httpbin
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.26 AS build
+FROM golang:1.26.4 AS build
 
 WORKDIR /go/src/github.com/mccutchen/go-httpbin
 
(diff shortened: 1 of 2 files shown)


ajithcnambiar:
@matheusfm could you please take a look, thanks! 🙏

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax = docker/dockerfile:1.3
-FROM golang:1.26 AS build
+FROM golang:1.26.4 AS build
```
mccutchen: I'm a little conflicted about this — it pins to the latest stable release _today_ but will quickly become a pin to an outdated stable release.

The choice of 1.26 was intentional, making it possible to track the latest stable release as it changes over time.  There are definitely tradeoffs, though, as the current approach might require explicitly running `docker pull golang:1.26` to make sure you're actually building on the latest stable release.

My inclination would be to drop this change, in part so that I don't have to keep updating it myself.

mccutchen (commented):
Yikes, thank you for the fix!

I've neglected this Dockerfile since #244, which uses [Dockerfile.release](https://github.com/mccutchen/go-httpbin/blob/main/Dockerfile.release) via goreleaser.  I need to figure out how to unify the two, this is a good reminder.

Dockerfile
ajithcnambiar: Thanks for the review, Ok - if you want me remove `Dockerfile` change and want to merge just `Makefile`, open to do that, thanks!

PR #268, opened 2026-08-12
feat: publish OpenAPI schema and embedded Swagger UI

### LINKED ISSUE #177: feat: openapi schema
As a user I'd like a swagger schema for the known APIs to be provided on /swagger.yml (or appropriate path) by the service itself.

- does such a schema exist?
- is there already a way to request it?

rayzorinc:
## Summary

go-httpbin can now describe and explore its HTTP API without relying on an external schema host. It serves an OpenAPI 3.2 document from `/openapi.json`, provides a fully embedded Swagger UI under `/swagger/`, and preserves both routes when the service runs with a path prefix.

The schema is derived from the same `http.ServeMux` patterns used at runtime, which keeps methods and path parameters aligned with the implementation. Swagger assets are vendored into the binary for offline and deployment reliability, and `/swagger/doc.json` provides a familiar schema alias.

A generated `docs/openapi.yaml` snapshot makes the schema available in the repository. `make openapi` refreshes it, while `make openapi-check` and `make lint` detect drift.

Fixes #177.

## Validation

- `make fmt lint test`
- Verified the prefixed routes locally at `/api/openapi.json`, `/api/swagger/doc.json`, and `/api/swagger/`.


mccutchen:
Raymond! It's been too long, I hope you're doing well!

A few high level thoughts on this change:
- Providing a complete and accurate OpenAPI spec is something I've wanted to do for a long time, but httpbin is a very weird kind of service where many of its endpoints/behaviors are specifically difficult to usefully describe in OpenAPI terms.  (I've got a long-lived WIP branch where I'm also trying to do this, I pick it up every 6 months or so before giving up again.)
  - The implementation here, which AFAICT from reading the code will generate a spec describing only the HTTP verb(s) and endpoints but missing any useful details about the parameters or other request/response behaviors feels a bit incomplete
- For now, I'm maintaining a strict zero dependencies policy
- Any OpenAPI implmenetation would necessarily need to be accompanied by an automated test suite to ensure that the OpenAPI spec and the actual implementation stay in sync over time (see, e.g., [schemathesis](https://schemathesis.readthedocs.io/en/stable/) for one way we might do that)
- Relatively minor and potentially short-sighted, but I'd probably start by loading Swagger from a CDN instead of committing it to the repo, until that became problematic

mccutchen:
(Apologies for accidental close!)

rayzorinc:
@mccutchen no worries.  seems like you have something in progress then and is actively working on it?  i can close mine then.

PR #270, opened 2026-08-25
feat: add /image/avif endpoint

### LINKED ISSUE #195: feat: add /image/avif endpoint
Add a endpoint for serving AVIF images at /image/avif

ChrisJr404:
Adds an `/image/avif` endpoint and wires AVIF into the Accept-based negotiation on `/image`, so a request with `Accept: image/avif` gets it back. Works the same way the existing png/jpeg/webp/svg formats do.

The asset is the same picture as the other image formats, just re-encoded to AVIF.

Closes #195.


mccutchen (approved):
👍 thanks!

